PR #13, opened 2015-11-21
Fixed structured data errors.

quangbahoa:
"Structured Data Testing Tool"


ThierryA:
Hey @quangbahoa,

Looking good! Could you please remove the space [after the not statement](https://github.com/quangbahoa/Beans/blame/master/lib/templates/structure/loop.php#L44). So change it to `!is_search()`. I am being picky but this is what is going to keep Beans clean and awesome under hood too!

You may submit a new pull request and I will merge it today so it will be in 1.2.2 which should be pushed next week :)

Thanks again for your contribution.

Issue #16, opened 2015-11-22
Wrong image path in Multisite Post Options metabox


gabormeszoly:
It looks like this:
```
http://beans-multisite.dev/site/wp-content/themes/tm-beans/lib/admin/assets/images/layouts/c.png
http://beans-multisite.dev/site/wp-content/themes/tm-beans/lib/admin/assets/images/layouts/cs.png
http://beans-multisite.dev/site/wp-content/themes/tm-beans/lib/admin/assets/images/layouts/sc.png
http://beans-multisite.dev/site/wp-content/themes/tm-beans/lib/admin/assets/images/layouts/css.png
http://beans-multisite.dev/site/wp-content/themes/tm-beans/lib/admin/assets/images/layouts/ssc.png
http://beans-multisite.dev/site/wp-content/themes/tm-beans/lib/admin/assets/images/layouts/scs.png
```

ThierryA:
Hey @gabormeszoly,

The paths looks correct to me, is your install in a `site` folder? What should the path be for that install?

Thanks,

gabormeszoly:
Hi @ThierryA 
I did a local version and everything is fine, probably my remote install has something wrong.
Sorry for that ...

ThierryA:
No worries, please do not hesitate to report if you can reproduce it :)

frankomator:
Hi @gabormeszoly, can you share your solution? I've got the same problem. Only the (correct) paths to the layout images are shown in the Customizer and in the Post Options but not the images themselves.
![screenshot](https://cloud.githubusercontent.com/assets/3018128/11335548/56f17ab4-91dd-11e5-9b49-0bd7f42ba77e.png)

ThierryA:
Alright guys, this is due to the `getimagesize()` which is not picking up that it is an image and therefore displaying the option as a radio button. I will push a fix and keep you updated so that you can do some testing.

Thanks,

gabormeszoly:
This problem comes up when the site is authorised. Example:
`Warning: getimagesize(http://congressline-multisite.dev.templatebooster.com/event2020/wp-content/themes/tm-beans/lib/admin/assets/images/layouts/c.png): failed to open stream: HTTP request failed! HTTP/1.1 401 Authorization Required in /var/www/dev.templatebooster.com/congressline-multisite/wp-content/themes/tm-beans/lib/api/fields/types/radio.php on line 38`

[...]

PR #17, opened 2015-11-23
Fixed getimagesize error when image is unavailable by url.

gabormeszoly:
It's a small but efficient fix from my colleague.


ThierryA:
Hi @gabormeszoly, thanks for the PR :)

This is unfortunately not the right approach as the field API is meant to be re-usable and the images are not going to be in the `BEANS_ADMIN_ASSETS_PATH ` when used by other devs.

I have push a fix 01ce2b8 and await for your response on #16.

Thanks,

PS: when submitting PR, make sure to pull from the master before to make sure you get all the latest changes and avoid conflicts.

PR #37, opened 2016-07-25
Modified beans_footer_credit_right and beans_footer_credit_left to div

gmmedia:
If we add a menu or something like that into one of the span elements, the result is a div element inside a span element. Div inside span is not valid.

### DIFF
--- a/lib/templates/fragments/footer.php
+++ b/lib/templates/fragments/footer.php
@@ -16,7 +16,7 @@ function beans_footer_content() {
 
 	beans_open_markup_e( 'beans_footer_credit', 'div', array( 'class' => 'uk-clearfix uk-text-small uk-text-muted' ) );
 
-		beans_open_markup_e( 'beans_footer_credit_left', 'span', array(
+		beans_open_markup_e( 'beans_footer_credit_left', 'div', array(
 			'class' => 'uk-align-medium-left uk-margin-small-bottom'
 		) );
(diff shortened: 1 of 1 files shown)


ThierryA:
Hey @gmmedia, big thanks for dropping by and suggest this change.

This is the nature of customization and we unfortunately can't cater for every possible way people would customize their theme. Instead, we strive to make the default welle coded and allow people to customize it according to their needs.

If you need to change these HTML markup to make it a `div`, you can just do:

```
beans_modify_markup( 'beans_footer_credit_left', 'div' );
beans_modify_markup( 'beans_footer_credit_right', 'div' );
```

Alternatively, you could even remove it if you prefer to only keep your customization by doing so:

```
beans_remove_markup( 'beans_footer_credit_left' );
beans_remove_markup( 'beans_footer_credit_right' );
```

Note that doing so **would not remove the hooks around the markup**, it would only remove it if you pass `true` as a second argument which I would **not advise** unless you really understand the consequences.

The other very important thing to keep in mind is **backwards compatibility**. If we change these `span` to `div`, all the current users who have styled their `span` would loose their styling which is definitely not something we would want to do **unless it is absolutely needed**. This is the nature of building products and we strive to make as stable as possible and avoid breaking sites on update ;-)

Once again, thanks for dropping by and feel free to submit other contribution!

PR #39, opened 2016-09-27
Fixed SiteNavigationElements name tag for microdata of schema

sminamie:


ThierryA:
Thanks for your submission @sminamie :-)

We want to prevent from changing the HTML as much as possible to avoid backwards compatibility issues. Adding the item name on the `li` is perfectly acceptable (more info [here](http://webmasters.stackexchange.com/questions/58926/schema-org-sitenavigationelement-markup/58944#58944).

However, if you prefer to have it as a span for your own needs, you can use the code snippet below in your child theme.

```
beans_remove_attribute( 'beans_menu_item_text', 'itemprop' );

beans_wrap_inner_markup( 'beans_menu_item_link', 'example_beans_menu_item_link_inner', 'span', array(
	'itemprop' => 'name',
) );
```

Happy coding,

sminamie:
Hi Thierry
Thank you for your advice.

Issue #52, opened 2017-11-06
Define our coding standards


hellofromtonya:
Create a set of coding standards to clearly communicate our code style for ourselves and anyone seeking to contribute to Beans.

Issue #46, opened 2017-11-06
Rework to be WPCS compliant.


hellofromtonya:
Rework the codebase to comply with WPCS.  At a minimum it should comply with WordPress-Extra standards setup and Travis.

ThierryA:
I would suggest:

```
<rule ref="WordPress">
	<exclude name="Generic.WhiteSpace.ScopeIndent.IncorrectExact" />
	<exclude name="Generic.WhiteSpace.ScopeIndent.Incorrect" />
</rule>
```

We might have to add a few other exclusions which I can't think of at this point (perhaps filename).

GaryJones:
I've got an initial `composer.json` to pull in some PHPCS standards, so I'll get that added as a PR in the next few days.

hellofromtonya:
@GaryJones has a few minor changes for PR #58.  Once we get that merged, then I'll start working on making the codebase compliant.

PR #58, opened 2017-11-11
Initial PHPCS setup

GaryJones:
This is some initial exploration with PHPCS for Beans. I'm not expecting any or all of this PR to be merged as is, but it can be used as a starting point for your discussions.

Comments have been added to key points of the PR.

See #46, #52.


composer.json
```diff
@@ -0,0 +1,32 @@
+{
```
GaryJones: Adding a `composer.json` means this project can then be added to Packagist, which then makes it considerably easier to pull it down as a dependency on a site. It also allows us to specify dependencies, which for here, relate to PHPCS.

composer.json
```diff
@@ -0,0 +1,32 @@
...
+    "name": "beans/beans",
```
GaryJones: I'm unsure what vendor name you want here.

composer.json
```diff
@@ -0,0 +1,32 @@
...
+    "description": "The most innovative theme for WordPress. The real magic is under the hood.",
+    "type": "wordpress-theme",
```
GaryJones: This is so that Composer Installers know where to move a theme a to, instead of inside the typical `vendor` directory.

composer.json
```diff
@@ -0,0 +1,32 @@
...
+    "license": "GPL-2.0+",
+    "homepage": "http://www.getbeans.io/",
```
GaryJones: This website seems to redirect `https://` to the `http://` site, which seems backwards.

composer.json
```diff
@@ -0,0 +1,32 @@
...
+    "support": {
+        "issues": "https://github.com/GetBeans/Beans/issues",
+        "source": "https://github.com/GetBeans/Beans"
+    },
+    "minimum-stability": "dev",
+    "prefer-stable": true,
+    "require": {
+        "php": "^5.2|^7",
+        "composer/installers": "^1.4",
+        "roave/security-advisories": "dev-master"
```
GaryJones: Recommended, so that packages with known security issues are flagged.

composer.json
```diff
@@ -0,0 +1,32 @@
...
+    },
+    "require-dev": {
+        "dealerdirect/phpcodesniffer-composer-installer": "^0.4",
```
GaryJones: Handles the registration of PHPCS code standards automatically after `composer install`.

composer.json
```diff
@@ -0,0 +1,32 @@
...
+        "sirbrillig/phpcs-variable-analysis": "^2.0",
```
GaryJones: Currently has a [bug](https://github.com/sirbrillig/phpcs-variable-analysis/issues/4) affecting 2.0 which means it's only half working, but I expect this to be addressed soon.

composer.json
```diff
@@ -0,0 +1,32 @@
...
+        "wimg/php-compatibility": "^8.0",
+        "wp-coding-standards/wpcs": "^0.14.0"
+    },
+    "config": {
+        "sort-order": true
+    },
+    "scripts": {
+        "install-codestandards": [
```
GaryJones: Only needed to help existing installs make use of the DeadlerDirect composer plugin.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0"?>
+<ruleset name="Beans">
+    <description>The code standard for Beans.</description>
+    <file>.</file>
+    <exclude-pattern>dist/*</exclude-pattern>
+    <exclude-pattern>node_modules/*</exclude-pattern>
```
GaryJones: May not be needed right now, but likely will be needed once other Node packages are added.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0"?>
+<ruleset name="Beans">
+    <description>The code standard for Beans.</description>
+    <file>.</file>
+    <exclude-pattern>dist/*</exclude-pattern>
```
GaryJones: Not sure if you plan to distribute this project as zip file, but if so, the build can be done to the de facto directory of `dist/`.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+    <exclude-pattern>node_modules/*</exclude-pattern>
+    <exclude-pattern>vendor/*</exclude-pattern>
+    <arg name="extensions" value="php"/>
```
GaryJones: Only check PHP for now - JS can be better checked with other tools.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+    <arg value="sp"/>
```
GaryJones: Show standards in PHPCS reports + show progress during PHPCS runs.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+    <arg name="report" value="full"/>
+    <arg name="report" value="summary"/>
+    <arg name="report" value="source"/>
```
GaryJones: Summary and source reports help to summarise the state of the project, beyond just the list of issues per file.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+    <rule ref="PHPCompatibility" />
+    <config name="testVersion" value="5.2-"/>
```
GaryJones: This can be bumped later on - currently flags up syntax that doesn't exist in the chosen version (i.e. short array syntax if still supporting PHP 5.2).

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+    <rule ref="WordPress">
+        <exclude name="WordPress.VIP"/>
+    </rule>
+    <rule ref="WordPress.Files.FileName">
```
GaryJones: The conditions here are typical - I tend to have classes named as per PSR-4, to allow considerably easier autoloading. It's a deviation from WP CS, but one which gives technical benefits through cleaner code and maintainability.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+        <properties>
+            <property name="is_theme" value="true"/>
+            <property name="strict_class_file_names" value="false"/>
+        </properties>
+        <exclude name="WordPress.Files.FileName.NotHyphenatedLowercase"/>
+    </rule>
+    <rule ref="WordPress.NamingConventions.PrefixAllGlobals">
+        <properties>
+            <property name="prefixes" type="array" value="beans,_beans,_Beans" />
```
GaryJones: This is a problem. The project has inconsistent prefixes for functions and classes (and doesn't use namespaces), and even with this group of prefixes, there are still global entities which don't use any prefix.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+        </properties>
+    </rule>
+    <rule ref="WordPress.WhiteSpace.ControlStructureSpacing">
+        <properties>
+            <property name="blank_line_check" value="true" />
```
GaryJones: Optional, but one I prefer. Flags (and can auto-fix) redundant blank lines, when indentation already highlights a change of context.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+        </properties>
+    </rule>
+    <rule ref="WordPress.WP.DeprecatedFunctions">
+        <properties>
+            <property name="minimum_supported_version" value="4.8" />
```
GaryJones: Not sure what you need to support, but the closer to the current release, the better to reduce the support burden.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+        </properties>
+    </rule>
+    <rule ref="WordPress.WP.I18n">
+        <properties>
+            <property name="text_domain" type="array" value="tm-beans" />
```
GaryJones: This is a problem. The text domain should match the theme slug (i.e. the directory it would be installed as under `wp-content/themes`. Presumably, this would be `beans`, so all of the text domain instances should be updated, and this value changed.

composer.json
hellofromtonya: We have multiple names:

- beans 
- tm-beans 
- getbeans 

Let’s standardize on beans.  

Therefore, let’s stick with vendor name of `beans` and the package name of `beans`.

composer.json
hellofromtonya: I’ll add that to my list of things to fix. Thank you.

phpcs.xml.dist
hellofromtonya: Yes, there will be other node modules as we bring in gulp, linting, babel, etc.

phpcs.xml.dist
hellofromtonya: These will all be addressed, some in the next release and then standardizing on our namespace when we bump up to PHP 5.6 in v2.

composer.json
hellofromtonya: Exactly.  Good job.

phpcs.xml.dist
hellofromtonya: @GaryJones You are correct.  In versions <2.0, beans _should be_ installed as `tm-beans` and not `beans` within the `wp-content/themes` folder. If not, any child themes that extend as `tm-beans` will get an error:

>### Broken Themes
>The following themes are installed but incomplete.

|**Name**   |    **Description**   |
|------------| -------------------|
|Beans child |  The parent theme is missing. Please install the "tm-beans" parent theme. |

## Mark as new issue

I opened [issue #59](https://github.com/Getbeans/Beans/issues/59) to propose changing the text domain and theme folder name to `beans` in version 2.0.

## Version < 2.0

In versions <2.0, we'll stick with `tm-beans`.  And yes, the theme's folder name is also `tm-beans`.

.gitignore
```diff
@@ -0,0 +1 @@
+/vendor/
```
hellofromtonya: Let's add `node_modules` too.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0"?>
+<ruleset name="Beans">
+    <description>The code standard for Beans.</description>
+    <file>.</file>
```
hellofromtonya: Add a new empty before this line of code.

phpcs.xml.dist
hellofromtonya: Add a new empty line and comment above this line of code, as we want to make this file more readable for all contributors:

```

    <!-- Ignore these folders -->
    <exclude-pattern>dist/*</exclude-pattern>
    <exclude-pattern>node_modules/*</exclude-pattern>
    <exclude-pattern>vendor/*</exclude-pattern>
```

phpcs.xml.dist
hellofromtonya: To make it more readable for all contributes, add a new empty line and comment:

```
    <!-- PHP only. We'll use other validators for JS, CSS, etc. -->
    <arg name="extensions" value="php"/>
```

phpcs.xml.dist
hellofromtonya: New empty line + comment

```

    <!-- Reporting Configuration -->
    <arg value="sp"/>
    <arg name="report" value="full"/>
    <arg name="report" value="summary"/>
    <arg name="report" value="source"/>
```

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+    <exclude-pattern>dist/*</exclude-pattern>
+    <exclude-pattern>node_modules/*</exclude-pattern>
+    <exclude-pattern>vendor/*</exclude-pattern>
+    <arg name="extensions" value="php"/>
+    <arg value="sp"/>
+    <arg name="report" value="full"/>
+    <arg name="report" value="summary"/>
+    <arg name="report" value="source"/>
+    <rule ref="PHPCompatibility" />
```
hellofromtonya: New empty line + comment

```

    <!-- PHP Rules -->
    <rule ref="PHPCompatibility"/>
    <config name="testVersion" value="5.2-"/>
```

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+    <config name="testVersion" value="5.2-"/>
+    <rule ref="WordPress">
```
hellofromtonya: New empty line + comment 

```

    <!-- WordPress Rules -->
    <rule ref="WordPress">
        <exclude name="WordPress.VIP"/>
    </rule>
```

And @GaryJones, let's add a space between each of the `<rule...></rule>` blocks too, which I'll note below.

phpcs.xml.dist
hellofromtonya: Add a new empty line above this `<rule>` to make it more readable.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+        <exclude name="WordPress.VIP"/>
+    </rule>
+    <rule ref="WordPress.Files.FileName">
+        <properties>
+            <property name="is_theme" value="true"/>
+            <property name="strict_class_file_names" value="false"/>
+        </properties>
+        <exclude name="WordPress.Files.FileName.NotHyphenatedLowercase"/>
+    </rule>
+    <rule ref="WordPress.NamingConventions.PrefixAllGlobals">
```
hellofromtonya: Add a new empty line above this `<rule>` to make it more readable.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+        <properties>
+            <property name="prefixes" type="array" value="beans,_beans,_Beans" />
+        </properties>
+    </rule>
+    <rule ref="WordPress.WhiteSpace.ControlStructureSpacing">
```
hellofromtonya: Add a new empty line above this `<rule>` to make it more readable.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+        <properties>
+            <property name="blank_line_check" value="true" />
+        </properties>
+    </rule>
+    <rule ref="WordPress.WP.DeprecatedFunctions">
```
hellofromtonya: Add a new empty line above this `<rule>` to make it more readable.

phpcs.xml.dist
```diff
@@ -0,0 +1,45 @@
...
+        <properties>
+            <property name="minimum_supported_version" value="4.8" />
+        </properties>
+    </rule>
+    <rule ref="WordPress.WP.I18n">
```
hellofromtonya: Add a new empty line above this `<rule>` to make it more readable.

GaryJones:
e4fd544 added which handles the requested changes.

hellofromtonya:
Thank you @GaryJones for contributing!  We appreciate your time and expertise!

Issue #47, opened 2017-11-06
Add unit tests


hellofromtonya:
1. Create tests for the codebase. 
2. Integrate CI with Travis.
3. Add badging.

PR #67, opened 2017-11-16
Made API Utilities WPCS Compliant

### LINKED ISSUE #63: Relative URLs Fail in beans_url_to_path
When passing a relative URL to `beans_url_to_path()`, the "Fix protocol" code does not recognize the URL.  As a result, the original URL passed to the function is appended as part of the final path.

For example:

```
example.com
```

is converted into 

```
{the value of ABSPATH}/example.com
```

when it should be just `ABSPATH`.

## Code in question

```
	if ( preg_match( '#^(\/\/)#', $url ) ) {
		$url = 'http:' . $url;
	}
```

## Proposed Solution

The proposed solution is to use `parse_url` instead to sniff out the URL's scheme.  If it's missing, we then fix the protocol. 

```
	if ( ! parse_url( $url, PHP_URL_SCHEME ) ) {
		$url = 'http://' . $url;
	}
```

### LINKED ISSUE #64: Non-internal URLs are being altered within `beans_url_to_path`
In `beans_url_to_path()`, which is within the `API\Utilities` package, if a URL is passed in that is not internal (meaning the URL Host is not the same), it's being passed through `beans_sanitize_path`.  

What happens?  It strips one of the protocol `//`. Whoops.

## For example

Let's say the site's URL is `http://getbeans.io` and that we want to convert `http://www.example.com/image.png` into an absolute path.  Since the image is not an internal URL, what gets returned is: `http:/www.example.com/image.png`.

## Current Code

Here is the current code:

```
	// Stop here if it is not an internal url.
	if ( false === stripos( $url, parse_url( site_url(), PHP_URL_HOST ) ) ) {
		return beans_sanitize_path( $url );
	}
```

## Solution

I propose that we simply return the `$url` instead of passing it into `beans_sanitize_path`.  In other words, if it's not an internal URL, send it back unchanged.

### LINKED ISSUE #65: An external URL with the site's root URL within doesn't bail out in `beans_url_to_path`
Although this is a wonky and probably rare edge case, an external URL that has the website's URL Host within it's path structure does not bail out.  I'm sure this needs more explanation.

Here is the current code:

```
	// Stop here if it is not an internal url.
	if ( false === stripos( $url, parse_url( site_url(), PHP_URL_HOST ) ) ) {
		return beans_sanitize_path( $url );
	}
```

Let's say the website's root URL is `https://getbeans.io`.  The external URL being passed into `beans_url_to_path()` might be `http://example.com/cool-stuff-at-getbeans.io`.  This URL is external as `example.com` is not the same host.

## Why?

`stripos()` searches the entire string looking for the first occurrence.  In the above edge case, it finds it.  But it's part of the path and not the host.

These test fail:

```
		Functions\expect( 'site_url' )->once()->andReturn( 'http://getbeans.io' );
		$this->assertSame(
			'http://example.com/cool-stuff-at-getbeans.io',
			beans_url_to_path( 'http://example.com/cool-stuff-at-getbeans.io' )
		);

		Functions\expect( 'site_url' )->once()->andReturn( 'https://www.getbeans.io' );
		$this->assertSame(
			'http://example.com/cool-stuff-at-www.getbeans.io',
			beans_url_to_path( 'http://example.com/cool-stuff-at-www.getbeans.io' )
		);

		Functions\expect( 'site_url' )->once()->andReturn( 'http://shop.getbeans.io' );
		$this->assertSame(
			'http://example.com/cool-stuff-at-shop.getbeans.io',
			beans_url_to_path( 'http://example.com/cool-stuff-at-shop.getbeans.io' )
		);
```

## Proposed Solution

We need to compare the hosts with one another.  We can do that like this:

```
	// It's not an internal URL. Bail out.
	if ( false === stripos( parse_url( $url, PHP_URL_HOST ), parse_url( $site_url, PHP_URL_HOST ) ) ) {
		return $url;
	}
```

BUT hold on.... if it's a relative URL, then the above won't work.  Why? `parse_url` won't process it.

Instead, let's move the protocol fix up, do it first, and then do the above comparison.  In doing so, we'll have to cache the original URL to return it.

```
	// Fix protocol. It isn't needed to set SSL as it is only used to parse the URL.
	if ( ! parse_url( $url, PHP_URL_SCHEME ) ) {
		$original_url = $url;
		$url          = 'http://' . ltrim( $url, '/' );
	}

	// It's not an internal URL. Bail out.
	if ( false === stripos( parse_url( $url, PHP_URL_HOST ), parse_url( $site_url, PHP_URL_HOST ) ) ) {
		return isset( $original_url ) ? $original_url : $url;
	}
```

All tests pass with the above solution.

### LINKED ISSUE #66: Remove tilde is not working in `beans_url_to_path`
For a URL with a tilde, such as `https://foo.com/~subdomain`, the tilde is not being removed.  

The following tests fail:

```
	public function testRemovesTilde() {
		Functions\expect( 'is_main_site' )->andReturn( true );
		Functions\expect( 'site_url' )->andReturn( 'https://foo.com' );

		$this->assertSame( rtrim( ABSPATH, '/' ), beans_url_to_path( 'https://foo.com/~subdomain', true ) );
		$this->assertSame( rtrim( ABSPATH, '/' ), beans_url_to_path( 'https://foo.com/~subdomain/', true ) );
		$this->assertSame( ABSPATH . 'foo', beans_url_to_path( 'https://foo.com/~subdomain/foo', true ) );
		$this->assertSame( ABSPATH . 'foo/', beans_url_to_path( 'https://foo.com/~subdomain/foo/', true ) );
	}
```

with the original code:

```
	$url  = parse_url( $url, PHP_URL_PATH );
	$path = wp_normalize_path( $url );
	$explode = beans_get( 0, explode( '/' , trailingslashit( ltrim( $path, '/' ) ) ) );

	// Maybe remove tilde from path.
	if ( false !== stripos( $explode, '~' ) ) {
		$path = preg_replace( '#\~[^/]*\/#', '', $path );
	}
```

## Proposed Solution

```
	// Parse url and standardize backslashes.
	$url  = parse_url( $url, PHP_URL_PATH );
	$path = wp_normalize_path( $url );

	// Maybe remove tilde from path.
	$trimmed_path = trailingslashit( ltrim( $path, '/' ) );
	$maybe_tilde  = beans_get( 0, explode( '/', $trimmed_path ) );
	if ( false !== stripos( $maybe_tilde, '~' ) ) {
		$ends_with_slash = substr( $path, - 1 ) === '/';
		$path            = preg_replace( '#\~[^/]*\/#', '', $trimmed_path );
		if ( $path && ! $ends_with_slash ) {
			$path = rtrim( $path, '/' );
		}
	}
```

The above solution uses the same search and regex.  However, it's using the trimmed path for the replace task.  Then if we still have a path and the original did not end with a `/`, we trim it.

All tests pass.

### LINKED ISSUE #68: Inaccurate Count for `beans_count_recursive` with `$count_parent` set to `true`
With the `$count_parent` set to `true`, the count is too high and doesn't correlate to the number of elements within the array.

For example, this array has 9 elements in it:

```
array( 1, 2, 3, 4, 5, array( 6, 7, 8, 9 ) )
```

When passing this array into `beans_count_recursive`, it returns a value of 15 elements.

Here is the test for it:

```
// The expect is 9. But the actual is 15.
$this->assertEquals( 9, beans_count_recursive( array( 1, 2, 3, 4, 5, array( 6, 7, 8, 9 ) ), 2 ) );
```

## Why? What is happening?

This line of code counts all of the elements in the array as the starting count:

```
	$count = $count_parent ? count( $array ) : 0;
```

Let's walk the counter:

1. On first pass, the `$count = 6`, as the base array has 6 elements in it.
2. Then we do the `foreach`.  For values `1-5`, the count increments.
3. When it gets to the 5th element, which is an array, the count value is now `$count = 11`.
4. The function is called again (recursive).
5. As the `$depth = 1` now, it counts the number of elements and returns `4`.
6. The `4` is added to the `11`.
7. When it's done, it gives us `15` elements.

## Works with `$parent_count` set to `false`

It works with the `$parent_count` set to `false`.  Why? The initial `$count` is initialized to a value of `0`.

```
// The expect is 9. The actual is 9.
$this->assertEquals( 9, beans_count_recursive( array( 1, 2, 3, 4, 5, array( 6, 7, 8, 9 ) ), 2, false ) );
```

hellofromtonya:
1. WPCS compliant see #46
2. Unit tests see #47
3. Fixes #63
4. Fixes #64
5. Fixes #65
6. Fixes #66
7. Closes #68 - deprecated the function.
8. Adds composer scripts.
9. Adds PHPUnit development suite setup.


composer.json
```diff
@@ -19,7 +29,12 @@
         "dealerdirect/phpcodesniffer-composer-installer": "^0.4",
         "sirbrillig/phpcs-variable-analysis": "^2.0",
         "wimg/php-compatibility": "^8.0",
-        "wp-coding-standards/wpcs": "^0.14.0"
+        "wp-coding-standards/wpcs": "^0.14.0",
+        "phpunit/phpunit": "^6.2",
+        "antecedent/patchwork": "2.0.8",
+        "brain/monkey": "^2.0",
+        "mockery/mockery": "~0.9",
+        "wpfulcrum/extender": "^3.0"
```
hellofromtonya: Adding all of the above for testing purposes.

lib/api/utilities/polyfills.php
```diff
@@ -0,0 +1,32 @@
...
+			if ( is_array( $value ) && is_array( $from_base = beans_get( $key, $base ) ) ) {
+				$base[ $key ] = array_replace_recursive( $from_base, $value );
+			} else {
+				$base[ $key ] = $value;
+			}
+		}
+
+		return $base;
+	}
+}
```
hellofromtonya: Moving the PHP 5.2 polyfill into its own file, where we can capture all polyfills.

phpcs.xml
```diff
@@ -21,10 +23,29 @@
...
+        <exclude name="Squiz.Commenting.FunctionComment.ExtraParamComment"/>
+        <!-- Will fix in v2 -->
+        <exclude name="PEAR.NamingConventions.ValidClassName.StartWithCapital"/>
+        <!-- Not every file needs a file header comment, e.g. Views -->
+        <exclude name="Squiz.Commenting.FileComment.Missing"/>
+        <!-- There are valid areas requiring errors to be silenced. -->
+        <exclude name="Generic.PHP.NoSilencedErrors.Discouraged"/>
+
+        <!-- Exclude until v2 as this applies after WP 4.4 with upgrades in PHP 5.3 -->
+        <exclude name="WordPress.WP.AlternativeFunctions.parse_url_parse_url"/>
```
hellofromtonya: Adjusted `phpcs.xml` for our current needs to be as compliant as possible.

tests/bootstrap.php
```diff
@@ -0,0 +1,33 @@
...
+ * @package     Beans\BeansFramework\Tests
+ * @since       1.5.0
+ * @link        http://www.getbeans.io
+ * @license     GNU-2.0+
+ */
+
+if ( version_compare( phpversion(), '5.3.0', '<' ) ) {
+	die( 'Beans Unit Tests require PHP 5.3 or higher.' );
+}
+
```
hellofromtonya: Although the codebase is at PHP 5.2, the testing tools are at 5.6.  Brain/Monkey is a good example.  The risk is minimal as we also do integration and full-on testing too (at PHP 5.2).

tests/bootstrap.php
```diff
@@ -0,0 +1,33 @@
...
+define( 'BEANS_TESTS_DIR', dirname( __FILE__ ) );
+
+define( 'BEANS_TESTS_LIB_DIR', BEANS_TESTS_DIR . '/../lib/' );
+
+/**
+ * You can turn on or off the integration tests.  Integration tests cause WordPress to load and run.
+ * PHP UnitTests, on the other hand, do not load WordPress, using Brain Monkey instead.
+ */
+define( 'BEANS_RUN_INTEGRATION_TESTS', false );
```
hellofromtonya: Tests in this package are currently Unit Tests and not integration tests.  Therefore, we do not load WordPress.  Instead, we're using Brain/Monkey and mocking out WordPress functionality.

tests/wp-integration-tests/bootstrap.php
```diff
@@ -0,0 +1,57 @@
+<?php
```
hellofromtonya: This file may change once we start building integration tests.

lib/api/utilities/functions.php
```diff
@@ -81,172 +81,173 @@ function beans_remove_dir( $dir_path ) {
...
- * @param string $path Path to be converted. Accepts absolute and relative internal paths.
+ * @param string $path          Path to be converted. Accepts absolute and relative internal paths.
+ * @param bool   $force_rebuild (Optional) Forces the rebuild of the root url and path.
  *
  * @return string Url.
  */
-function beans_path_to_url( $path ) {
-
-	static $root, $host;
+function beans_path_to_url( $path, $force_rebuild = false ) {
```
hellofromtonya: For testing purposes, I added a `force_rebuild` parameter.  There may also be times where you want to force a rebuild.  One use case that comes to mind is:

In a multisite, I've had situations where I need to programmatically switch to a different site, grab some information and process, and then switch back.  In this case, I would need the ability to clear the cache and rebuild.

lib/api/utilities/functions.php
```diff
@@ -81,172 +81,173 @@ function beans_remove_dir( $dir_path ) {
...
+	static $root_path, $root_url;
```
hellofromtonya: Changed the variable names to be more clear as to the values they represent.

lib/api/utilities/functions.php
```diff
@@ -81,172 +81,173 @@ function beans_remove_dir( $dir_path ) {
...
+			$root_url  = preg_replace( $pattern, '', $root_url );
+		}
 
-			// Add the blog path for multsites.
-			if ( ! is_main_site() && ( $blogdetails = get_blog_details( get_current_blog_id() ) ) ) {
-				if ( ! ( defined( 'WP_SITEURL' ) ) || ( defined( 'WP_SITEURL' ) && WP_SITEURL == site_url() ) ) {
-					$host = untrailingslashit( $host ) . $blogdetails->path;
-				}
+		// If it's a multisite and not the main site, then add the site's path.
+		if ( ! is_main_site() ) {
```
hellofromtonya: Doing the multisite work is not a requirement of the subfolder task.  They are 2 separate tasks.  So I unwrapped it from the subfolder removal section.

lib/api/utilities/functions.php
```diff
@@ -81,172 +81,173 @@ function beans_remove_dir( $dir_path ) {
...
+ * @param bool   $force_rebuild (Optional) Forces the rebuild of the root url and path.
  *
  * @return string Absolute path.
  */
-function beans_url_to_path( $url ) {
-
-	static $root, $blogdetails;
+function beans_url_to_path( $url, $force_rebuild = false ) {
+	static $root_path, $blogdetails;
+	$site_url = site_url();
```
hellofromtonya: We hit `site_url()` a couple of times.  That function has a lot going on.  Therefore, let's cache it here and then reuse the variable for a tiny little optimization bump.

lib/api/utilities/functions.php
```diff
@@ -81,172 +81,173 @@ function beans_remove_dir( $dir_path ) {
...
-		if ( ! is_main_site() ) {
-
-			// Set blogdetails if it isn't cached.
-			if ( ! $blogdetails ) {
-				$blogdetails = get_blog_details( get_current_blog_id() );
-			}
-
-			$path = preg_replace( '#^(\/?)' . trailingslashit( preg_quote( ltrim( $blogdetails->path, '/' ) ) ) . '#', '', $path );
+	// Remove the blog path for multisites.
+	if ( ! is_main_site() ) {
```
hellofromtonya: Just like in the other function, I moved the multi-site task out of the subfolder task. See note in `beans_path_to_url`

hellofromtonya:
@ThierryA As we discussed, this PR does the following things for 1 file:

1. Makes the API\Utilities WPCS compliant.  I also adjust the rules too for our needs.
2. Adds unit tests.
3. During the testing process, I found a few issues.  I noted them in Issues and then tagged them within both this commit and the comments.  
4. All tests pass now.

Upon approval, we can use the process and structure done here for the remaining work within the codebase.  Step-by-step, it'll be easier to systematically repeat the process for each inner folder, moving outward until the theme is done.

## The Process

1. Build the test for a function.
2. Then make that function compliant.
3. Fix any bugs.
4. Repeat for the next function.

## A note on Tests:

I'm doing a series of tests including:

- live testing in single site and multi-site
- true unit tests using [`Brain\Monkey`](https://github.com/Brain-WP/BrainMonkey)
- then later we'll do integration tests with WordPress
- Sniffers automated within PhpStorm (boy is that nice!)

The unit testing environment is not at PHP 5.2.  Rather, I'm using tools that need PHP 5.6.  The risk here is small. Why? The multi-tiered testing approach applies us to ensure everything works at the 5.2 level.

## Refactoring

Some refactoring was done in this PR.  However, there are further improvements that can be made to reduce redundant code, type hint for arrays, etc.  I'll do that in the next PR.

lib/api/utilities/functions.php
```diff
@@ -81,172 +81,173 @@ function beans_remove_dir( $dir_path ) {
...
-	// Remove subfolder if necessary.
-	if ( '' !== ( $subfolder = parse_url( site_url(), PHP_URL_PATH ) ) ) {
-
-		// Set root if it isn't cached.
-		if ( isset( $set_root ) ) {
-
-			// Remove subfolder.
-			$root = preg_replace( '#' . untrailingslashit( preg_quote( $subfolder ) ) . '$#', '', $root );
+	// If the subfolder exists for the root URL, then strip it off of the root path.
+	// Why? We don't want a double subfolder in the final path.
```
ThierryA: This doesn't comply to the WP standards multi-line doc (see [here](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#5-2-multi-line-comments)).

lib/api/utilities/functions.php
```diff
@@ -6,22 +6,26 @@
...
  * Calls function given by the first parameter and passes the remaining parameters as arguments.
  *
  * The main purpose of this function is to store the content echoed by a function in a variable.
  *
  * @since 1.0.0
  *
- * @param callback $callback The callback to be called.
- * @param mixed    $var      Additional parameters to be passed to the callback.
+ * @param Callable $callback The callback to be called.
+ * @param mixed    $args,... (Optional) Additional parameters to be passed to the callback.
```
ThierryA: `(Optional)` format doesn't comply to WP code standards (see [here](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-functions-class-methods)) which may be important for the automatic parser. It should be `Optional.`

NB: this comment applies throughout this PR.

lib/api/utilities/functions.php
```diff
@@ -81,172 +81,173 @@ function beans_remove_dir( $dir_path ) {
...
+		$root_path = wp_normalize_path( untrailingslashit( ABSPATH ) );
+		$root_url  = untrailingslashit( site_url() );
 
 		// Remove subfolder if necessary.
-		if ( '' !== ( $subfolder = parse_url( $host, PHP_URL_PATH ) ) ) {
-
-			$root = preg_replace( '#' . untrailingslashit( preg_quote( $subfolder ) ) . '$#', '', $root );
-			$host = preg_replace( '#' . untrailingslashit( preg_quote( $subfolder ) ) . '$#', '', $host );
+		$subfolder = parse_url( $root_url, PHP_URL_PATH );
+		if ( $subfolder && '/' !== $subfolder ) {
```
ThierryA: I am in favor of having a blank before all statements (example `if`, `foreach` etc.) as part of our coding standards.

NB: this comment applies throughout this PR.

lib/api/utilities/polyfills.php
```diff
@@ -0,0 +1,32 @@
...
+ * @license     GNU-2.0+
+ */
+
+if ( ! function_exists( 'array_replace_recursive' ) ) {
+	/**
+	 * PHP 5.2 fallback.
+	 *
+	 * @ignore
+	 */
+	function array_replace_recursive( $base, $replacements ) {
```
ThierryA: Since we are moving towards WPCS compliancy, let's complete the PHPDoc here.

lib/api/utilities/functions.php
```diff
@@ -6,22 +6,26 @@
  * loaded automatically when the Beans framework is included.
  *
  * @package API\Utilities
+ *
+ * @since   1.5.0
  */
 
+// Load the dependencies.
+require_once dirname( __FILE__ ) . '/polyfills.php';
```
ThierryA: Let's keept the loaders together by moving this one [over there](https://github.com/Getbeans/Beans/blob/development/lib/api/init.php#L36).

lib/api/utilities/functions.php
hellofromtonya: Okay.  Let me reformat and adjust the WPCS ruleset.

hellofromtonya: @ThierryA In this particular case though, adding a new line breaks the grouping:

```
		// Remove subfolder if necessary.
		$subfolder = parse_url( $root_url, PHP_URL_PATH );

		if ( $subfolder && '/' !== $subfolder ) {
			$pattern   = '#' . untrailingslashit( preg_quote( $subfolder ) ) . '$#';
			$root_path = preg_replace( $pattern, '', $root_path );
			$root_url  = preg_replace( $pattern, '', $root_url );
		}
```

as all of the lines here go together. And the general comment above it applies to the group.

Do you prefer to keep things _grouped together_? Or stick with the general formatting ruleset for `BlankLineBeforeControlStructure`?

ThierryA: I would be in favor of stick to `BlankLineBeforeControlStructure` which is easier to follow for contributors and sniffers, removing the human judgement of whether it is acceptable not to have a blank line in certain situation or not :-)

hellofromtonya:
>It would be great to add Travis config in this PR (or separate and merge latest here) too so that we have WPCS and phpunit running from now on.

Let's do a separate PR just for configuring Travis.  It's easier to focus our code reviews when we break things up into smaller, more focused tasks.

hellofromtonya:
@ThierryA Commit 2 includes all changes for reformatting, WPCS exclude ruleset, etc. from your review above.

tests/bootstrap.php
GaryJones: If the testing tools require PHP 5.6, shouldn't the above message reflect that?

hellofromtonya:
>It would be great to add Travis config in this PR (or separate and merge latest here) too so that we have WPCS and phpunit running from now on.

Commit #3 adds Travis CI.  For now it is only configured for the tests and not WPCS.  Why? WPCS would fail.  Once we get the codebase compliant, then we can add the configurations to Travis. :)

lib/api/utilities/functions.php
GaryJones: AFAIK, WPCS (sniffs) doesn't check for this `Optional.` format specifically, but it could do in the future, since it is in the Handbook, and if it's being added, you may as well add it in the format that Core does.

lib/api/utilities/functions.php
GaryJones: FWIW, I'm in favour of ThierryA's preferred format.

> And the general comment above it applies to the group.

This is just a code smell that the code could be refactored out later on, so I wouldn't worry about assigning a value to a variable, and the first instance of using that variable being together (since later uses would also likely not be grouped together as well anyway).

lib/api/utilities/functions.php
```diff
@@ -81,172 +81,173 @@ function beans_remove_dir( $dir_path ) {
...
+	if ( true === $force_rebuild ) {
+		$root_path   = '';
+		$blogdetails = '';
+	}
 
-	// Stop here if it is not an internal url.
-	if ( false === stripos( $url, parse_url( site_url(), PHP_URL_HOST ) ) ) {
-		return beans_sanitize_path( $url );
+	// Fix protocol. It isn't needed to set SSL as it is only used to parse the URL.
+	// Issue #63. Fixed.
```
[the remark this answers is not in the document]
GaryJones: Ticket numbers mean little but full URLs can definitely make it easier to read-up on why code was written in a certain way - the [`@see`](http://docs.phpdoc.org/references/phpdoc/tags/see.html) (for DocBlocks) exists for that very reason.

.travis.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    include:
+        - php: 5.2
+          dist: precise
+        - php: 5.6
+        - php: 7.0
+        - php: 7.1
+
+    allow_failures:
+        - php: 7.0
```
GaryJones: Why should PHP 7.1 be allowed to fail?

lib/api/utilities/functions.php
ThierryA: @GaryJones the entire [Code Reference](http://www.getbeans.io/code-reference/) is automatically created/maintained using WP Parser, I am not referring to WPCS ;-)

.travis.yml
```diff
@@ -0,0 +1,43 @@
...
+    - WP_VERSION=master WP_MULTISITE=0
+
+matrix:
+    fast_finish: true
+
+    include:
+        - php: 5.2
+          dist: precise
+        - php: 5.6
+        - php: 7.0
```
GaryJones: There's probably not much point in testing 7.0 and 7.1.

`7.1` (or `7.2` as of a couple of weeks time) + nightly (with allowed failure for nightly) would probably be sufficient.

composer.json
```diff
@@ -19,14 +29,21 @@
         "dealerdirect/phpcodesniffer-composer-installer": "^0.4",
```
GaryJones: This plugin has a couple of gotcha bugs in 0.4.0 to 0.4.2, so I'd be more specific here and state `^0.4.3`. 

`0.4.3` or later almost certainly what will typically be installed, but it's one less potential source of issue to deal with.

composer.json
```diff
@@ -19,14 +29,21 @@
...
         "sirbrillig/phpcs-variable-analysis": "^2.0",
         "wimg/php-compatibility": "^8.0",
-        "wp-coding-standards/wpcs": "^0.14.0"
+        "wp-coding-standards/wpcs": "^0.14.0",
+        "phpunit/phpunit": "^6.2",
+        "antecedent/patchwork": "2.0.8",
```
GaryJones: This seems to be a rather specific version. I think `2.0.9` is out, but this won't allow for that. Is that intentional?
Any reason not to let the patchwork dependency in Brain Monkey handle this?

lib/api/utilities/functions.php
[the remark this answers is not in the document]
ThierryA: I agree only if it is to apply DRY principals and if it really can't be avoid, but if the code is written in a certain way then a comment should explain why instead for making it hard for the devs by linking to an external url which is not always guarantied to be accessible for life (and not accessible offline). Github is awesome today, but we don't want to marry the code base with it ;-)

composer.json
```diff
@@ -19,14 +29,21 @@
...
     },
     "config": {
         "sort-order": true
     },
     "scripts": {
         "install-codestandards": [
             "Dealerdirect\\Composer\\Plugin\\Installers\\PHPCodeSniffer\\Plugin::run"
-        ]
+        ],
+        "cs-check": "vendor/bin/phpcs --colors",
```
GaryJones: This colors arg can be defined in the phpcs.xml.dist. See https://github.com/squizlabs/PHP_CodeSniffer/blob/9b6260b4ff9410682fd34a34e68e377e60dc4961/phpcs.xml.dist#L13

composer.json
```diff
@@ -19,14 +29,21 @@
...
+        "test": "vendor/bin/phpunit --colors"
```
GaryJones: This colors arg can be specified in the `phpunit.xml.dist` file. See https://github.com/GaryJones/plugin-boilerplate/blob/bd9ca02455f5fc2f3b57ade0459d672b9472d392/phpunit.xml.dist#L10

tests/bootstrap.php
hellofromtonya: @GaryJones You're right.  Fixing in the next commit.

.travis.yml
hellofromtonya: @GaryJones Are you thinking no `allow_failures`?

.travis.yml
hellofromtonya: @GaryJones Good point, changing to PHP 7.2.  I've also commented out 5.2.  Yes, the codebase is built on 5.2, but our tests run at 5.6.

composer.json
hellofromtonya: @GaryJones It's a carry-over from your last PR.  I'm updating it now.  Pushing shortly.

composer.json
hellofromtonya: @GaryJones Good points. Removed to let Brain Monkey handle its own dependencies.  Pushing shortly.

phpcs.xml
```diff
@@ -5,9 +5,11 @@
     <file>.</file>
```
GaryJones: Why did this whole file get renamed?

The use of `phpcs.xml.dist` as the distributed version, and adding `/phpcs.xml` to `.gitignore`, means that developers can override aspects of `phpcs.xml.dist` with their local-only version of `phpcs.xml`.

(Note, that as of PHPCS 3.1.0, you can now use `.phpcs.xml.dist` and `.phpcs.xml` (with leading dot).)

phpcs.xml
```diff
@@ -5,9 +5,11 @@
...
 
     <!-- Ignore these directories -->
-    <exclude-pattern>dist/*</exclude-pattern>
+    <exclude-pattern>*/dist/*</exclude-pattern>
     <exclude-pattern>node_modules/*</exclude-pattern>
-    <exclude-pattern>vendor/*</exclude-pattern>
+    <exclude-pattern>*/tests/*</exclude-pattern>
+    <exclude-pattern>*/vendor/*</exclude-pattern>
```
GaryJones: Having both a `vendor` and ` vendors` directory is going to get confusing...

phpcs.xml
hellofromtonya: @GaryJones Because we are debating integrating XMP's dev_lib suite into the repo, though I'm leaning more towards [GrumPHP](https://github.com/phpro/grumphp).  If we go with XWP's suite, it has `phpcs.xml` in it. We want ours to be the master file.

ThierryA:
> Commit #3 adds Travis CI. For now it is only configured for the tests and not WPCS. Why? WPCS would fail. Once we get the codebase compliant, then we can add the configurations to Travis. :)

@hellofromtonya we should apply WPCS sniffs only for patches on Travis until the entire code base is compliant.

phpcs.xml
```diff
@@ -21,10 +23,31 @@
...
     <rule ref="WordPress">
         <exclude name="WordPress.VIP"/>
+        <!-- In the API, we process the requests. Nonce verification is to be done
+            at the form processing level prior to calling our API. -->
+        <exclude name="WordPress.CSRF.NonceVerification.NoNonceVerification"/>
+        <exclude name="WordPress.PHP.StrictInArray.MissingTrueStrict"/>
+        <!-- We have functions that can receive more arguments than declared
+            in the function's parameter list. -->
+        <exclude name="Squiz.Commenting.FunctionComment.ExtraParamComment"/>
+        <!-- Not every file needs a file header comment, e.g. Views -->
```
GaryJones: I'd suggest that, for consistency, and for the sake of not turning this whole check off for the whole project, that views do have a file-level DocBlock. Here's an [example](https://github.com/GaryJones/plugin-boilerplate/blob/bd9ca02455f5fc2f3b57ade0459d672b9472d392/phpunit.xml.dist#L10). Note that one could also add in namespaces, strict_type declarations and so on, as well as making it clear to anyone opening the view file out of context, what project it is from and where it is used.

phpcs.xml
hellofromtonya: @GaryJones We have folders in the codebase that are named `vendor`.  We'll correct the naming to standardize in v2.

phpcs.xml
ThierryA: @hellofromtonya we can choose the file name with the dev-lib by setting the `export WPCS_STANDARD={filename}.xml` in a `.dev-lib` file.

phpcs.xml
```diff
@@ -21,10 +23,31 @@
...
+        <exclude name="Squiz.Commenting.FileComment.Missing"/>
+        <!-- There are valid areas requiring errors to be silenced. -->
+        <exclude name="Generic.PHP.NoSilencedErrors.Discouraged"/>
```
GaryJones: As of PHPCS 3.2.0, you can [selectively ignore specific sniffs on specific lines](https://github.com/squizlabs/PHP_CodeSniffer/blob/8a18d4049dda4349cfe01dffc1420bbce31d304e/package.xml#L47), so this may be a better approach, rather than disabling the whole check, which means that other silenced errors may be accidentally introduced.

phpcs.xml
hellofromtonya: @GaryJones @ThierryA Renamed it to `phpcs.xml.dist`.  Update coming....

phpunit.xml.dist
```diff
@@ -0,0 +1,35 @@
+<phpunit
```
GaryJones: Adding in the full [XML references](https://github.com/GaryJones/plugin-boilerplate/blob/bd9ca02455f5fc2f3b57ade0459d672b9472d392/phpunit.xml.dist#L1-L3) (update URL to the right version), allows IDEs to pull the XSD and check the XML in this file is valid.

phpunit.xml.dist
```diff
@@ -0,0 +1,35 @@
...
+        <directory suffix=".php">./tests/</directory>
+    </testsuite>
+
+    <filter>
+        <whitelist processUncoveredFilesFromWhitelist="true">
+            <directory suffix=".php">.</directory>
+        </whitelist>
+    </filter>
+
+    <logging>
```
GaryJones: That's a lot of logging. Is it really all needed?

phpcs.xml
hellofromtonya: @GaryJones With this new change in 3.2.0, is this done within the `phpcs.xml` file or as an annotation?

tests/unit-tests/api/utilities/A_RenderFunctionTest.php
```diff
@@ -0,0 +1,91 @@
...
+			'foo, bar, baz',
+			beans_render_function( __NAMESPACE__ . '\renderFunctionCallback', array( 'foo', 'bar' ), 'baz' )
+		);
+	}
+
+	/**************************************
+	 * Helpers
+	 *************************************/
+
+	public function renderFunctionCallback() {
```
GaryJones: Why not mock these callback functions, instead of actually defining them?

phpunit.xml.dist
hellofromtonya: @GaryJones My bad...that was for my internal testing during the setup process.  Stripping it out everything by the `coverage-text`.

tests/unit-tests/api/utilities/C_RemoveDirTest.php
```diff
@@ -0,0 +1,105 @@
...
+		if ( is_dir( $this->dir ) ) {
+			if ( ! is_writable( $this->dir ) ) {
+				chmod( $this->dir, 0755 );
+			}
+			rmdir( $this->dir );
+		}
+	}
+
+	public function testBailsForNonDir() {
+		$dir = dirname( __FILE__ ) . '/';
```
GaryJones: This can be `__DIR__`.

tests/unit-tests/api/utilities/A_RenderFunctionTest.php
hellofromtonya: @GaryJones I wanted to ensure that the callback is truly called and that the arguments are being passed.  While we can mock it out, you'll see that I'm doing some processing work to further test the expected behavior.

tests/unit-tests/api/utilities/C_RemoveDirTest.php
hellofromtonya: @GaryJones Yes it can. `__DIR__` and `dirname(__FILE__)` are [equivalent to one another](http://php.net/manual/en/language.constants.predefined.php).

.travis.yml
GaryJones: Allow failures on `nightly`, maybe HHVM, if you care to test for that (I wouldn't bother).

See https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/02da0322dab24ba65bf1483136b375dc9706f21c/.travis.yml#L53-L56

composer.json
GaryJones: If we can't find fault in our own code, then we'll have transcended to a higher level of coding ability :-)

phpcs.xml
GaryJones: Per line comment. i.e.:
```php
@ini_set .... // phpcs:disable Generic.PHP.NoSilencedErrors.Discouraged
```
(or something similar)

hellofromtonya: @GaryJones Implemented.  But upon running we are getting Warnings:

```
 90 | WARNING | Silencing errors is discouraged (Generic.PHP.NoSilencedErrors.Discouraged)
 94 | WARNING | Silencing errors is discouraged (Generic.PHP.NoSilencedErrors.Discouraged)
```

lib/api/utilities/functions.php
```diff
@@ -86,11 +86,11 @@ function beans_remove_dir( $dir_path ) {
 		if ( is_dir( $path ) ) {
 			beans_remove_dir( $path );
 		} else {
-			@unlink( $path );
+			@unlink( $path ); // phpcs:disable Generic.PHP.NoSilencedErrors.Discouraged - This is a valid use case.
```
GaryJones: Just to note, this requires PHPCS 3.2.0, which is not released yet, but probably will be soon. I don't know if the other standards will have any fixing to do to be able to use it.

hellofromtonya:
@ThierryA Here's where we stand right now.  Some Composer `require-dev` package is requiring PHP 7.1.  That is making Travis CI fail:

[5.6](https://travis-ci.org/Getbeans/Beans/jobs/303555649)

[7.0](https://travis-ci.org/Getbeans/Beans/jobs/303555650)

I'll take some time to sniff out what package is causing the problem.  You may see some commits that seem silly.  But they will be needed to test on Travis.

In hindsight, I should have separated Travis into a separate PR as this PR is doing too many things.

ThierryA:
Sure @hellofromtonya, we should break things into different PRs and commits as much as we can (to a certain extent indeed). I would say that we should revert the changes for Travis on this PR and open another PR for it. I think Travis should be a high priority so to make the CR easier, quicker and avoid missing things (we are only humans).

hellofromtonya:
@ThierryA 

> I would say that we should revert the changes for Travis on this PR and open another PR for it. I think Travis should be a high priority so to make the CR easier, quicker and avoid missing things (we are only humans).

I completely agree.  Let me strip it out of this PR.  Then once this one is merged in, my next priority is getting Travis online and functionality per our conversations.

hellofromtonya:
Per our conversation, I've renamed all of the unit tests to comply with the WordPress Unit Test Coding Standard instead of PSR-4.

GaryJones:
> I'll take some time to sniff out what package is causing the problem.

`composer why php`

lib/api/utilities/functions.php
```diff
@@ -81,172 +81,181 @@ function beans_remove_dir( $dir_path ) {
...
+	if ( true === $force_rebuild ) {
+		$root_path   = '';
+		$blogdetails = '';
+	}
 
-	// Stop here if it is not an internal url.
-	if ( false === stripos( $url, parse_url( site_url(), PHP_URL_HOST ) ) ) {
-		return beans_sanitize_path( $url );
+	// Fix protocol. It isn't needed to set SSL as it is only used to parse the URL.
+	if ( ! parse_url( $url, PHP_URL_SCHEME ) ) {
```
ThierryA: We must confirm (test) that this change works with IP addresses too.

tests/unit-tests/api/utilities/Tests_Functions.php
```diff
@@ -0,0 +1,497 @@
...
+		Functions\expect( 'is_main_site' )->andReturn( true );
+		Functions\expect( 'site_url' )->andReturn( 'https://foo.com' );
+
+		$this->assertSame( rtrim( ABSPATH, '/' ), beans_url_to_path( 'https://foo.com/~subdomain', true ) );
+		$this->assertSame( rtrim( ABSPATH, '/' ), beans_url_to_path( 'https://foo.com/~subdomain/', true ) );
+		$this->assertSame( ABSPATH . 'foo', beans_url_to_path( 'https://foo.com/~subdomain/foo', true ) );
+		$this->assertSame( ABSPATH . 'foo/', beans_url_to_path( 'https://foo.com/~subdomain/foo/', true ) );
+
+		// @TODO-Check with Thierry on this behavior.
+		$this->assertSame( ABSPATH . 'bar/~subdomain/foo/', beans_url_to_path( 'https://foo.com/bar/~subdomain/foo/', true ) );
```
ThierryA: Yes this would be the expected behavior.

tests/unit-tests/api/utilities/Tests_Functions.php
```diff
@@ -0,0 +1,497 @@
+<?php
```
ThierryA: Domains are used in multiple tests throughout this file, IP address should be tested as well.

ThierryA: [This comment](https://github.com/Getbeans/Beans/pull/67#discussion_r151621237) applies to this file too.

ThierryA: Add simple PHPDoc for class and methods in this file to replace non WP standards compliant comments. Example:

```
/***********************************************
 * Tests for beans_render_function()
 **********************************************/
```

I would be in favor of using the PHPUnit `@covers` tag too (see [here](https://phpunit.de/manual/3.7/en/appendixes.annotations.html#appendixes.annotations.covers)).

ThierryA: This file isn't 100% WPCS compliant. Whitelisting comments may be used where need be (example `echo` may be tagged with `// WPCS: XSS ok.`

tests/bootstrap.php
ThierryA: I would advise not to run unit tests and integration tests together. We could use the phpunit `--testsuite` (or else) selection option to run integration tests separately.

tests/unit-tests/api/utilities/Tests_Functions_RemoveDir.php
```diff
@@ -0,0 +1,105 @@
+<?php
```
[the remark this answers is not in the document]
hellofromtonya: These are ina separate file because of the clean up task that occurs after each test.  Why run these cleanups for all of the other tests when they do not apply?

hellofromtonya:
Namespaces are named as follows:

`VendorName\PackageName`

Beans is the vendor name.  
BeansFramework is the package.

When we build plugins in the near future, we’ll name those `Beans\PluginName`. 

The vendor name of Beans ties everything together, making it relatable and quickly identifiable to the Beans suite of products.

Also think about if someone else picks that single namespace name of `Beans`.  Right?  The entire point of namepsacing is to combine enough unique names together to make the entire fully qualified name unique.

tests/unit-tests/api/utilities/Tests_Functions.php
hellofromtonya: I often use this technique to provide a grouping comment for like types of tests, methods, functionality, etc.  It is not tied to a single method, but rather a group of methods.  It provides a quick visual to say "Hey, this type of comment means everything from here down ....or until the next one like this....are grouped together."

phpcs.xml.dist
```diff
@@ -21,10 +24,27 @@
     <!-- PHP Rules -->
     <rule ref="PHPCompatibility" />
     <config name="testVersion" value="5.2-"/>
+    <rule ref="Squiz.PHP.DisallowMultipleAssignments.Found">
```
GaryJones: What's the particular use case for this being excluded?

phpcs.xml.dist
```diff
@@ -21,10 +24,27 @@
...
+        <severity>0</severity>
+    </rule>
 
     <!-- WordPress Rules -->
     <rule ref="WordPress">
         <exclude name="WordPress.VIP"/>
+        <!-- In the API, we process the requests. Nonce verification is to be done
+            at the form processing level prior to calling our API. -->
+        <exclude name="WordPress.CSRF.NonceVerification.NoNonceVerification"/>
+        <exclude name="WordPress.PHP.StrictInArray.MissingTrueStrict"/>
```
GaryJones: Is this to be addressed later?

phpcs.xml.dist
```diff
@@ -45,6 +65,9 @@
         <properties>
             <property name="blank_line_check" value="true" />
         </properties>
+        <!-- Beans requires a new blank line BEFORE each control structure.
+             Excluding this rule will allow a blank line after all control structures. -->
+        <exclude name="WordPress.WhiteSpace.ControlStructureSpacing.BlankLineAfterStart"/>
```
GaryJones: I don't believe these three lines are in sync with each other.

The `BlankLineAfterStart` check aims to remove the following blank line _inside_ the structure:

```php
if ( true ) {

    do something();
}
```

As such, it has no bearing on new lines _before_ a control structure (line before `if`), or after a control structure (line after closing `}`), which is what the two comment lines refer to.

phpunit.xml.dist
```diff
@@ -0,0 +1,29 @@
...
+         beStrictAboutCoversAnnotation="true"
+         beStrictAboutOutputDuringTests="true"
+         beStrictAboutTestsThatDoNotTestAnything="true"
+         beStrictAboutTodoAnnotatedTests="true"
+         convertErrorsToExceptions="true"
+         convertNoticesToExceptions="true"
+         convertWarningsToExceptions="true"
+         verbose="true">
+
+    <testsuite name="beans">
```
GaryJones: How about making this `unit`, or `integration`?

tests/bootstrap.php
GaryJones: Example of a [possible directory setup](https://github.com/GaryJones/plugin-boilerplate/tree/master/tests), including the `bootstrap.php` which conditionally loads WP depending on the `--testsuite` that is run.

tests/unit-tests/api/utilities/Tests_Functions_RemoveDir.php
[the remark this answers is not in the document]
GaryJones: The code here appears to be actually interacting with the real file system.

I'd rather see vfsStream being used here - see https://github.com/mikey179/vfsStream/wiki/Example as a good introduction.

tests/unit-tests/polyfills/functions.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
```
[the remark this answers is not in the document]
GaryJones: > I am in favor of forcing Beans to be in a WP environment for pure Unit Tests though.

Unless you've missed a "not" in your sentence, this would appear to be contradictory. Unit tests should be runnable without any DB, any file system, and any other dependencies like WP.

GaryJones:
I'm all for making a project follow CS, but for me, this PR still does too much:

- PHPCS config
- PHPUnit config
- production code
- test code (incl new files)
- discussion on namespaces, class and file naming, new unit tests etc.

If the aim is to follow CS, then get the `phpcs.xml.dist` agreed and committed, and then work on individual directories (cherry pick from here into new PRs) to make it bite size.

Discussions on new tests, choice of namespaces, tags in file-level DocBlocks and so on should be in other PRs IMHO.

GaryJones:
FWIW, I think `Beans\BeansFramework` or `Beans\Framework` would be fine - there doesn't appear to be any existing package on Packagist that has the `Beans` vendor name, so the shorter package name would be fine. It really depends on if the package is known as "the Beans Framework" or "the framework by Beans".

Also, it may be worth considering changing the repo name (`github.com/GetBeans/BeansFramework` or `github.com/GetBeans/Framework` to match), or even contacting GitHub, and asking if `github.com/beans` account could be freed up to be come an organisation.

lib/api/utilities/functions.php
hellofromtonya: Confirmed.  New commit adds IP addresses to tests.

hellofromtonya:
@ThierryA @GaryJones All of your comments were addressed in the latest 2 commits.

1. IP address testing is added.
2. File renaming for WPCS compliance.
3. DocBlock for test methods and class.
4. Using vsStream for filesystem testing.
5. Reworked the PHPUnit testsuite bootstrapping using a Listener.
6. Test suites are separated:

To run the unit tests, do the following:
```
phpunit --testsuite unit
```

To run the integration tests, do the following:
```
phpunit --testsuite integration
```

lib/api/utilities/functions.php
```diff
@@ -375,7 +374,7 @@ function beans_count_recursive( $array, $depth = false, $count_parent = true ) {
 		return count( $array, COUNT_RECURSIVE );
 	}
 
-	$count = $count_parent ? count( $array ) : 0;
+	$count = $count_parent ? count($array) : 0;
```
GaryJones: This seems like it goes against WPCS?

tests/integration-tests/class-test-case.php
```diff
@@ -1,32 +1,27 @@
...
 	/**
 	 * Prepares the test environment before each test.
-	 *
-	 * @since 1.0.0
-	 *
-	 * @return void
 	 */
 	protected function setUp() {
 		parent::setUp();
 		Monkey\setUp();
```
GaryJones: It's possible that we don't need to mock anything WP-specific in integration tests, since WP will be available, so either Mockery or mocks in PHPUit could be used.

In which case, this whole class would not be needed, and integration tests could extend `WP_UnitTestCase` directly.

hellofromtonya: @GaryJones It's possible.  Once we start writing those tests, then we can adjust this Test Case, if needed.

lib/api/utilities/functions.php
hellofromtonya: Yup.  Pushed a fix for it.  There's also a question that I'm working with @ThierryA  on to understand the intent of `$count_parent`.  But we can do that in the next PR.

hellofromtonya:
@ThierryA @GaryJones Let's discuss Test file naming and organization to ensure we are fully compliant to WPCS.

## What's in a single test class

Per the 1st paragraph in [WPCS Test Classes](https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#test-classes):

>As a rule, a single test class should not contain tests for more than one function/method.

Following this rule, our `tests-functions.php` file should be broken up into separate files, one test class per function.

I actually like this rule, as it makes working within a test file much easier. Why? You can find the test faster via the file name.  And once in the file, you know that everything in that file is for that single function/method.

This is how I originally had it and how I'm used to writing tests.  

## File naming

@ThierryA  noted above:
>Rename file to tests-functions.php.

But let's look at the WordPress Writing PHPUnit Tests handbook to compare.

> Test files should be named using camelCase.

Test class names should reflect the filepath, with underscores replacing directory separators and TitleCase replacing camelCase. Thus, the class `Tests_Comment_GetCommentClass()`, which contains tests for the `get_comment_class()` function, is located in `tests/comment/getCommentClass.php`.

## Rework to Comply

Given the entire [Test Classes specification](https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#test-classes), that means we should:

1. Break apart the `tests-functions.php` file into separate files, as noted above.
2. Then name each of those test files.  We'll have these when done: 
```
tests/api/utilities/beansRenderFunction.php
tests/api/utilities/beansRenderFunctionArray.php
tests/api/utilities/beansRemoveDir.php
tests/api/utilities/beansPathToUrl.php
tests/api/utilities/beansSanitizePath.php
tests/api/utilities/beansUrlToPath.php
tests/api/utilities/beansGet.php
tests/api/utilities/beansMultiArrayKeyExists.php
tests/api/utilities/beansCountRecursive.php
```

@GaryJones @ThierryA  Let's vote.  Should we break it apart and rename per the specification?  I say Yes, as our goal is to be fully compliant.

hellofromtonya:
I rebased the last 4 commits to roll them up for final review.  Let's talk about what's here.

## List of What's In this Commit

1. Renamed files, classes, and methods to be WPCS compliant (more on that below)
2. Added method comments
3. Added tests for IP address
4. Corrected beans_count_recursive tests (though [Issue #68]() remains open (next PR).
5. Added vfsStream for filesystem testing
6. Separated the test suites (see below).

## Separated Test Suites

Per @ThierryA request, the test suites have been separated.  To run the unit tests, do the following:
```
phpunit --testsuite unit
```

To run the integration tests, do the following:
```
phpunit --testsuite integration
```
For the integration tests, we need to load the bootstrap file.  I built a Bootstrap Listener which will load that file when it's an integration suite.  There's a setup test case in the integration file which serves the simple purpose of letting PHPUnit know "Hey, there are tests in here."

## WPCS Compliant

After reading the WPCS Testing Handbook, the way we were naming files, classes, and methods was wrong.

I know we want to be fully compliant.  Therefore, I went ahead and transformed our tests to comply.

### What's in a single test class

Per the 1st paragraph in [WPCS Test Classes](https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#test-classes):

>As a rule, a single test class should not contain tests for more than one function/method.

Following this rule, our `tests-functions.php` file should be broken up into separate files, one test class per function.

I actually like this rule, as it makes working within a test file much easier. Why? You can find the test faster via the file name.  And once in the file, you know that everything in that file is for that single function/method.

### File naming

@ThierryA  noted above:
>Rename file to tests-functions.php.

But let's look at the WordPress Writing PHPUnit Tests handbook to compare.

> Test files should be named using camelCase.
>
>Test class names should reflect the filepath, with underscores replacing directory separators and TitleCase replacing camelCase. Thus, the class `Tests_Comment_GetCommentClass()`, which contains tests for the `get_comment_class()` function, is located in `tests/comment/getCommentClass.php`.

### Reworked to Comply

Given the entire [Test Classes specification](https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#test-classes), that means we should:

1. Break apart the `tests-functions.php` file into separate files, as noted above.
2. Then name each of those test files.  We'll have these when done: 
```
tests/api/utilities/beansRenderFunction.php
tests/api/utilities/beansRenderFunctionArray.php
tests/api/utilities/beansRemoveDir.php
tests/api/utilities/beansPathToUrl.php
tests/api/utilities/beansSanitizePath.php
tests/api/utilities/beansUrlToPath.php
tests/api/utilities/beansGet.php
tests/api/utilities/beansMultiArrayKeyExists.php
tests/api/utilities/beansCountRecursive.php
```

I did that in this commit.

ThierryA:
@hellofromtonya thanks for the research and last commit. Even though I don't fully agree with the PHPUnit doc which goes against some standards, I think that it is important that we fully comply to Core. Green light from me there :-)

phpcs.xml.dist
```diff
@@ -21,10 +23,27 @@
...
+    <rule ref="Squiz.PHP.DisallowMultipleAssignments.Found">
+        <severity>0</severity>
+    </rule>
 
     <!-- WordPress Rules -->
     <rule ref="WordPress">
         <exclude name="WordPress.VIP"/>
+        <!-- In the API, we process the requests. Nonce verification is to be done
+            at the form processing level prior to calling our API. -->
+        <exclude name="WordPress.CSRF.NonceVerification.NoNonceVerification"/>
```
ThierryA: I would prefer adding the exclude at the file line level using whitelisting comment, for example:
``` PHP
return $_POST; // phpcs:disable WordPress.CSRF.NonceVerification.NoNonceVerification for X reason.
```
or it my be a bit less specific
``` PHP
return $_POST; // @codingStandardsIgnoreLine for X reason.
```

phpcs.xml.dist
```diff
@@ -21,10 +23,27 @@
...
+        <exclude name="WordPress.PHP.StrictInArray.MissingTrueStrict"/>
+
+        <!-- There are valid areas requiring errors to be silenced.
+             Remove when PHPCS 3.2.0 is released. -->
+        <exclude name="Generic.PHP.NoSilencedErrors.Discouraged"/>
```
ThierryA: I think this may be removed.

phpcs.xml.dist
```diff
@@ -21,10 +23,27 @@
...
+
+        <!-- Runs that we'll fix in v2 -->
+
+        <!-- Exclude until v2 as this applies after WP 4.4 with upgrades in PHP 5.3 -->
+        <exclude name="WordPress.WP.AlternativeFunctions.parse_url_parse_url"/>
```
ThierryA: [This comment](https://github.com/Getbeans/Beans/pull/67/files#r152414338) applies here too.

tests/unit-tests/polyfills/functions.php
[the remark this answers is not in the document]
ThierryA: @GaryJones I did miss "not" in my sentence indeed ;-)

tests/unit-tests/api/utilities/beansCountRecursive.php
```diff
@@ -0,0 +1,83 @@
...
+			array( 1, 2, ),
+			3,
+			array(
+				4,
+				5,
+				array( 6, 7, 8, 9, array( 10, 11 ) ),
+			),
+		), 4, false ) );
+
+		// This test fails.  See Issue #68
```
ThierryA: Kindly refer to [this reply](https://github.com/Getbeans/Beans/issues/68#issuecomment-346181947).

README.md
```diff
@@ -1,4 +1,7 @@
 # Beans WordPress theme framework
+
+[![Build Status](https://travis-ci.org/Getbeans/Beans.svg?branch=development)](https://travis-ci.org/Getbeans/Beans)
```
ThierryA: Remove this for now since it will be applied in a separate PR.

tests/unit-tests/api/utilities/beansPathToUrl.php
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	/**
+	 * Test beans_path_to_url() should bail out when relative URL.
+	 */
+	public function test_should_bail_out_when_relative_url() {
+		$this->assertSame( '//getbeans.io', beans_path_to_url( '//getbeans.io' ) );
+		$this->assertSame( '//www.getbeans.io', beans_path_to_url( '//www.getbeans.io' ) );
+
+//		$this->assertSame( 'getbeans.io', beans_path_to_url( 'getbeans.io' ) );
```
ThierryA: Why are these commented out?

tests/unit-tests/api/utilities/beansPathToUrl.php
```diff
@@ -0,0 +1,338 @@
...
+	 */
+	public function test_should_convert_to_ip_when_absolute_path() {
+		Functions\expect( 'is_main_site' )->andReturn( true );
+
+		$url = 'https://8.8.8.8';
+		Functions\expect( 'site_url' )->andReturn( $url );
+
+		$this->assertSame( "{$url}/api/utilities/beansPathToUrl.php", beans_path_to_url( __FILE__, true ) );
+
+		$path = __DIR__;
```
ThierryA: I think this might be a leftover.

ThierryA:
@hellofromtonya we are very close to getting this merged :smile: In addition to my last comments above, the `unit-tests` files don't fully pass the WPCS sniffs, mostly due to:

- `Squiz.Commenting.FileComment.Missing`
- `WordPress.Arrays.ArrayIndentation.ItemNotAligned`
- `WordPress.Arrays.ArrayDeclarationSpacing.AssociativeArrayFound`
- `WordPress.XSS.EscapeOutput.OutputNotEscaped`

and a few others.

Thanks @hellofromtonya, you absolutely rock!

phpcs.xml.dist
hellofromtonya: But that's not available yet until PHPCS 2.3 or something is released.  Once that's released, then we can use comment annotations.

tests/unit-tests/api/utilities/beansCountRecursive.php
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Class Tests_Beans_Count_Recursive
+ *
+ * @package Beans\Framework\Tests\API\Utilities
+ *
+ * @group   unit-tests
+ * @group   api
+ */
+class Tests_Beans_Count_Recursive extends Test_Case {
```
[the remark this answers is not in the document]
hellofromtonya: Wow, confusing:

>Test class names should reflect the filepath, with underscores replacing directory separators and TitleCase replacing camelCase. Thus, the class Tests_Comment_GetCommentClass(), which contains tests for the get_comment_class() function, is located in tests/comment/getCommentClass.php.

But you're right.   Change is coming.

tests/unit-tests/api/utilities/beansPathToUrl.php
hellofromtonya: These 2 tests fail.  I was thinking about whether it needs to pass for these or not.  I left them in to remind me.  

Why do they fail?  We're using a `preg_match()` and looking for `data`, `//`, `http`, `https`, to let us know that this is a URL.  A relative URL without any of those is not caught.  Instead, it thinks it's a filesystem path.

In another PR, I'll be submitting some refactored code that moves redundant checks into private functions:

- checking for a tilde
- checking for subdirectory
- checking if it's a URL or a filesystem path
- etc.

During that PR, I'll address these 2 lines of code.  If I remember them, they may slip through the cracks.

tests/unit-tests/api/utilities/beansCountRecursive.php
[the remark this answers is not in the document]
ThierryA: Very confusing!

_We are dedicated to be Core compliant so let it be, but I don't like the Unit Tests standards to be honest ssssh 🤐_

phpcs.xml.dist
ThierryA: As far as I am aware, `// @codingStandardsIgnoreFile`, `// @codingStandardsIgnoreStart`, `// @codingStandardsIgnoreEnd` and `// @codingStandardsIgnoreLine` are so we can use that until PHPCS 3.2.0 is released.

hellofromtonya:
@ThierryA Whew, I think we've got it now.  The latest commit for review includes the following:

1. Moved the tests into `tests/phpunit` to match WordPress structure.  It is noted in WPCS.  This sets up the ability to add qunit tests.
2. Renamed the Test classes to be WPCS compliant.
3. Deprecated `beans_count_recursive()`.
4. Removed tests for `beans_count_recursive()`.
5. Reformatted and added comments for the tests to comply with phpcs sniffs.
6. Added inline comment phpcs test disables where appropriate.
7. Removed needed phpcs exclude tests from XML. (using inline comments instead).

tests/unit-tests/api/utilities/beansCountRecursive.php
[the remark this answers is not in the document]
GaryJones: Worth noting that the info about tests applies to WordPress Core in the first instance.

The WPCS, while in the Core Handbook as there's no where else for it, is intended to apply to all WP-related software, of which WP Core happens to be just one.

WP Core doesn't really distinguish between unit and integration tests for instance.

If Beans does, and it makes more sense to follow WPCS naming conventions so the whole project is consistent, then so be it.

If you look at https://core.trac.wordpress.org/attachment/ticket/41057/phpcs8.xml.dist which is the latest proposed patch for `phpcs.xml.dist` for WP Core, you'll see (neat the bottom) that they exclude the test files from file naming conventions. i.e., you can apply code standards for the contents of the test files, without feeling constrained about also matching file names.

lib/api/utilities/functions.php
```diff
@@ -294,7 +294,7 @@ function beans_sanitize_path( $path ) {
 function beans_get( $needle, $haystack = false, $default = null ) {
 
 	if ( false === $haystack ) {
-		$haystack = $_GET;
+		$haystack = $_GET; // phpcs:disable WordPress.CSRF.NonceVerification.NoNonceVerification as the nonce verification check should be at the form processing level.
```
GaryJones: PHPCS 3.2.0 and above.

phpcs.xml.dist
```diff
@@ -21,25 +21,14 @@
     <arg name="report" value="source"/>
 
     <!-- PHP Rules -->
-    <rule ref="PHPCompatibility" />
-    <config name="testVersion" value="5.2-"/>
-    <rule ref="Squiz.PHP.DisallowMultipleAssignments.Found">
-        <severity>0</severity>
+    <rule ref="PHPCompatibility">
+        <exclude-pattern>*/tests/*</exclude-pattern>
```
GaryJones: Worth adding a comment as to why this is excluded. 
If it's because the tests have a higher minimum version of PHP, then you can take one of two approaches.

1. Add a `phpcs.xml.dist` file inside the the tests directory, then add a second call to phpcs the composer script referencing this config file
2. Just add a runtime config value to the composer script. See [this](https://github.com/GaryJones/plugin-boilerplate/blob/6a0515058bbb59826b3be190f1930a2dea016982/composer.json#L58) as an example, which happens to test that my main plugin file is 5.2 compatible, but you can easily see how you could just check that your tests/ directory is 5.* compatible.

hellofromtonya:
@GaryJones @ThierryA I looked at what @GaryJones noted in the previous commit and made the following changes for PHPCS.

1. Added a note explaining why tests are excluded from the PHPCompatibility rules:

```
    <!-- PHP Rules -->
    <rule ref="PHPCompatibility">
        <!-- Tests are at PHP 5.6. -->
        <exclude-pattern>*/tests/*</exclude-pattern>
    </rule>
```
This will be removed in Beans 2.

2. Improved the Composer script naming and added a separate `phpcs-tests` script too:

```
    "phpcs": "vendor/bin/phpcs",
    "phpcs-tests": "vendor/bin/phpcs --runtime-set testVersion 5.6 tests/phpunit/",
    "test-unit": "vendor/bin/phpunit --testsuite unit",
    "test-integration": "vendor/bin/phpunit --testsuite integration"
```

**Note**

1. We want to retain `phpcs.xml.dist` PHP version for the IDEs to pick it up.  For example, in PhpStorm, the new changes are automatically recognized for me.  I can quickly see when I'm not compliant.
2. We could do a more complex setup as [suggested here](https://github.com/wimg/PHPCompatibility/issues/491).  I'm thinking though that's overkill, since we will quickly move to PHP 5.6 in Beans 2.

composer.json
```diff
@@ -41,7 +41,8 @@
     "install-codestandards": [
       "Dealerdirect\\Composer\\Plugin\\Installers\\PHPCodeSniffer\\Plugin::run"
     ],
-    "cs-check": "vendor/bin/phpcs",
+    "phpcs": "vendor/bin/phpcs",
```
GaryJones: By all means keep the individual scripts, but I'd rather see a script that ran both of these PHPCS together, otherwise it's going to be easy to miss one or the other.
```json
"phpcs-src": "vendor/bin/phpcs",
"phpcs-tests": "vendor/bin/phpcs --runtime-set testVersion 5.6 tests/phpunit/",
"phpcs": [
  "@phpcs-src",
  "@phpcs-tests"
],
```

Now all code can be checked with `composer phpcs`.

tests/phpunit/integration-tests/bootstrap.php
```diff
@@ -0,0 +1,59 @@
...
+ * @package     Beans\Framework\Tests
+ * @since       1.5.0
+ * @link        http://www.getbeans.io
+ * @license     GNU-2.0+
+ *
+ * @group       integrationtests
+ */
+
+// Require patchwork early so that functions can be monkey patched in Unit tests.
+require __DIR__ . '/../../vendor/antecedent/patchwork/Patchwork.php';
```
ThierryA: This need to look one level up now that it has moved.

tests/phpunit/class-bootstrap.php
```diff
@@ -0,0 +1,34 @@
...
+	/**
+	 * Start the test suite.
+	 *
+	 * @param \PHPUnit_Framework_TestSuite $suite Test suite instance.
+	 */
+	public function startTestSuite( PHPUnit_Framework_TestSuite $suite ) {
+
+		// Load the integration testsuite's bootstrap file.
+		if ( 'integration' === $suite->getName() ) {
+			require __DIR__ . '/integration-tests/bootstrap.php';
```
ThierryA: Change to `require_once` to prevent error as the listener runs multiple times.

tests/phpunit/integration-tests/bootstrap.php
```diff
@@ -0,0 +1,59 @@
...
+ * Bootstraps the WordPress Integration Tests
+ *
+ * @package     Beans\Framework\Tests
+ * @since       1.5.0
+ * @link        http://www.getbeans.io
+ * @license     GNU-2.0+
+ *
+ * @group       integrationtests
+ */
+
```
ThierryA: Return if we are not in a WP environment to avoid error.

ThierryA:
@hellofromtonya thanks for the latest changes, please see my last comments. This can be merged after that 🎉

PR #75, opened 2017-11-25
Fixed tests, dev env, and composer scripts for Windows

### LINKED ISSUE #71: composer.json scripts paths for phpcs and phpunit don't work on Windows.
The paths would need backslashes (like the "install-codestandards" path) to work on Windows.

```
   "phpcs-src": "vendor\\bin\\phpcs",
    "phpcs-tests": "vendor\\bin\\phpcs --runtime-set testVersion 5.6 tests\\phpunit\\",
    "phpcs": [
      "@phpcs-src",
      "@phpcs-tests"
    ],
    "test-unit": "vendor\\bin\\phpunit --testsuite unit",
    "test-integration": "vendor\\bin\\phpunit --testsuite integration"
```
I can submit a PR if appropriate.

hellofromtonya:
- Adds Windows-friendly filesystem paths for composer scripts
- Adds .editorconfig
- Changes directory constants to use DIRECTOR_SEPARATOR
- Fixes API Utilities tests for Windows filesystem
- Tested on OSX and Windows 10
- Closes #71
- Loads either PHPUnit 4.8 or 5.7


tests/phpunit/bootstrap.php
```diff
@@ -13,7 +13,8 @@
 }
 
 define( 'BEANS_TESTS_DIR', __DIR__ );
-define( 'BEANS_TESTS_LIB_DIR', dirname( dirname( __DIR__ ) ) . '/lib/' );
+define( 'BEANS_ROOT_DIR', dirname( dirname( __DIR__ ) ) . DIRECTORY_SEPARATOR );
+define( 'BEANS_TESTS_LIB_DIR', BEANS_ROOT_DIR . 'lib' . DIRECTORY_SEPARATOR );
```
hellofromtonya: All test directory constants will use the `DIRECTORY_SEPARATOR` to ensure tests work on both Windows and Mac environments.

composer.json
```diff
@@ -41,13 +41,13 @@
...
+    "phpcs-src": "\"vendor/bin/phpcs\"",
+    "phpcs-tests": "\"vendor/bin/phpcs\" --runtime-set testVersion 5.6 tests/phpunit/",
     "phpcs": [
       "@phpcs-src",
       "@phpcs-tests"
     ],
-    "test-unit": "vendor/bin/phpunit --testsuite unit",
-    "test-integration": "vendor/bin/phpunit --testsuite integration"
+    "test-unit": "\"vendor/bin/phpunit\" --testsuite unit",
+    "test-integration": "\"vendor/bin/phpunit\" --testsuite integration"
```
hellofromtonya: The backslashes are for Windows machines.  These changes work on both Mac and Windows.

tests/phpunit/unit-tests/api/utilities/beansSanitizePath.php
```diff
@@ -29,21 +29,102 @@ protected function setUp() {
...
 	}
 
+	/**
+	 * Test beans_sanitize_path() should remove Windows drive.
+	 */
+	public function test_should_remove_windows_drive() {
+		$this->assertSame( '/Program Files/tmp/', beans_sanitize_path( 'C:\Program Files\tmp\\' ) );
+		$this->assertSame( '/Foo/bar/baz/index.txt', beans_sanitize_path( 'D:\Foo\bar\baz\index.txt' ) );
+	}
+
```
hellofromtonya: Check first that it removes the Window's drive from the path before doing the rest of the tests.

tests/phpunit/unit-tests/api/utilities/beansSanitizePath.php
```diff
@@ -29,21 +29,102 @@ protected function setUp() {
...
+		// test beans' theme root path.
+		$this->assertSame(
+			$this->prepare_path( rtrim( BEANS_ROOT_DIR, DIRECTORY_SEPARATOR ) ),
+			beans_sanitize_path( BEANS_ROOT_DIR )
+		);
+		$directory_separator = '\\' === DIRECTORY_SEPARATOR ? '/' : '';
+		$this->assertSame(
+			$this->prepare_path( BEANS_ROOT_DIR ) . $directory_separator . 'functions.php',
+			beans_sanitize_path( BEANS_ROOT_DIR . 'functions.php' )
+		);
```
hellofromtonya: Changes here are:

1. using the `prepare_path()` method for the expected value.
2. using `DIRECTORY_SEPARATOR` where needed.

On a Mac, the `prepare_path()` method does nothing.  But on a Windows machine, `C:\Users\Tonya\` is changed to `/Users/Tonya/`.  We have some work to do to prepare the expected path.

tests/phpunit/unit-tests/api/utilities/beansSanitizePath.php
```diff
@@ -29,21 +29,102 @@ protected function setUp() {
...
 	}
 
 	/**
-	 * Test beans_sanitize_path() should remove Windows drive.
+	 * Test beans_sanitize_path() should sanitize for filesystem symbolic links.
 	 */
-	public function test_should_remove_windows_drive() {
-		$this->assertSame( '/Program Files/tmp/', beans_sanitize_path( 'C:\Program Files\tmp\\' ) );
-		$this->assertSame( '/Foo/bar/baz/index.txt', beans_sanitize_path( 'D:\Foo\bar\baz\index.txt' ) );
+	public function test_should_sanitize_for_filesystem_symbolic_links() {
```
hellofromtonya: The `beans_sanitize_path` function also converts the symbolic links of `/../` into an absolute path.  Let's test that separately and purposefully to ensure it works on Mac and Windows.

tests/phpunit/unit-tests/api/utilities/beansSanitizePath.php
```diff
@@ -29,21 +29,102 @@ protected function setUp() {
...
+
+		// It's not a Windows path. Just return.
+		if ( '\\' !== DIRECTORY_SEPARATOR ) {
+			return $path;
+		}
+
+		$path = wp_normalize_path( realpath( $path ) );
+
+		// Strips off the Windows drive letter.
+		return mb_substr( $path, 2 );
```
hellofromtonya: The `prepare_path()` does not nothing on a Mac or Linux machine.  But on a Windows machine, it has to do the following work for the expected path:

1. Convert it to a real absolute path.
2. Normalize it.
3. Strip of the drive, e.g. `C:`, `D:`, etc.

tests/phpunit/unit/api/utilities/beansSanitizePath.php
```diff
@@ -67,43 +67,24 @@ public function test_should_sanitize_for_filesystem() {
 	 * Test beans_sanitize_path() should sanitize for filesystem symbolic links.
 	 */
 	public function test_should_sanitize_for_filesystem_symbolic_links() {
```
ThierryA: @hellofromtonya I removed the tests for integration directory as they are completely separated now (`BEANS_TESTS_DIR` refers to the Unit directory and no longer the root).

tests/phpunit/unit/api/utilities/beansUrlToPath.php
```diff
@@ -21,13 +21,23 @@
...
+	protected static $abspath;
+
 	/**
 	 * Setup test fixture.
 	 */
 	protected function setUp() {
 		parent::setUp();
 
 		require_once BEANS_TESTS_LIB_DIR . 'api/utilities/functions.php';
+		if ( is_null( self::$abspath ) ) {
```
ThierryA: @hellofromtonya now that the `ABSPATH` is sanitized globally here, this shouldn't be needed and all `self::$abspath` should be replaced back to `ABSPATH`.

PS: I haven't tested it but it should be the case.

hellofromtonya: @ThierryA Actually no.  We still need that. Why?  The `ABSPATH` is the real fully qualified path on the computer's filesystem.  In our API, we need to get file access for our conversions as well as loading files, such as in the Compiler.  

In the conversions, as shown above, we need to strip off the Windows drive, e.g. C:, D:, E:, etc.  

Looking through the API, we need to retain the true `ABSPATH`, as it is in WordPress, meaning pointing to the root path of the website filesystem.  Why? We are loading a file from WordPress in the Compiler.  Plus, in the future, we may need to use `ABSPATH` for other purposes for filesystem access.

Therefore, I'm going to change `ABSPATH` to be exactly what it is in WordPress, instead of referencing the tests directory as I originally set it up.  That will save us a bunch of time and frustrations down the road.

I'm adjusting the tests now and will push the new commit.

hellofromtonya:
@ThierryA In the last commit, here's what I've done:

- Made `ABSPATH` the same as it is in WordPress (see above for reasons why). 
- Moved defining `ABSPATH` to the unit test's bootstrap file as it's a bootstrapping process.
- Fixed a missing space in the `composer test-integration` script path.
- Added relative pathing static properties in the `beansPathToUrl.php` setup.
- Tested on Mac, Windows, and TravisCI. All tests pass.

## ABSPATH

I think it's important to leave `ABSPATH` to the expected value and then adjust our tests accordingly.  Why? To ensure that when we need to know the true `ABSPATH`, our code and tests work as expected.  

For example, right now, we use `ABSPATH` in the Compiler to load a WordPress file.  

In the future, we may need its true value.  Therefore, it's better to set it up that way and then adjust our tests.

hellofromtonya:
@ThierryA This commit is ready for your final approval.

tests/phpunit/unit/api/utilities/beansPathToUrl.php
```diff
@@ -21,13 +21,37 @@
  */
 class Tests_BeansPathToUrl extends Test_Case {
 
+	/**
+	 * Relative path to the Beans directory starting from `wp-content`.
+	 *
+	 * @var string
+	 */
+	protected static $beans_relative_path;
```
ThierryA: Why making this static?

PS: this comment applies to `$beans_tests_relative_path` and `$abspath` in the `Tests_BeansUrlToPath`?

hellofromtonya: I typically use `static` on properties that don't change, though really these should be constants.  Regardless, it's a carry-over from working on other projects.  I've removed it and switched to protected properties.

ThierryA:
@hellofromtonya I haven't done actual testing on Windows and I will leave that up to you. The code looks good to go 👍

PR #85, opened 2017-12-06
Made `_beans_render_action` and `_beans_unique_action_id` compliant

### LINKED ISSUE #78: When passing more than 1 subhook to `_beans_render_action`, the 1st subhook is called 2x
When calling `_beans_render_action();` with more than one subhook, the first subhook is called twice.  That's not the expected behavior.

For example, when calling `_beans_render_action( 'foo[bar][baz]' );`, the call stack is:

1. `do_action( 'foo' );` fires.
2. `do_action( 'foo[bar]' );` fires.
3. `do_action( 'foo[baz]' );` fires.
4. `do_action( 'foo[bar]' );` fires again. <- Whoops, this shouldn't happen.
5. `do_action( 'foo[bar][baz]' );` fires.

In the above example, `do_action( 'foo[bar]' );` fires 2x.  It's happening in the [cascading sub-hooks section](https://github.com/Getbeans/Beans/blob/development/lib/api/actions/functions.php#L576).

## Tests

This test fails as the 1st hook is being called twice and not once:

```
	/**
	 * Test _beans_render_action() renders two levels of sub-hooks,
	 * but the original hook is not rendered.
	 */
	public function test_should_render_two_levels_of_sub_hooks_but_not_original() {
		$stub    = Beans_Api_Actions_Stub::class;
		$hook    = 'foo[bar][baz]';
		$message = 'Called me. ';

		// The root hook renders.
		add_action( 'foo', array( $stub, 'echo_static' ) );
		Actions\expectDone( 'foo' )->once()->whenHappen( array( $stub, 'echo_static' ) );
		$expected = $message;

		// The 1st sub-hook renders.
		add_action( 'foo[bar]', array( $stub, 'echo_static' ) );
		Actions\expectDone( 'foo[bar]' )->once()->whenHappen( array( $stub, 'echo_static' ) );
		$expected .= $message . $message;

		// These hooks will not render as they are not registered.
		Actions\expectDone( 'foo[baz]' )->never();
		Actions\expectDone( 'foo[bar][baz]' )->never();

		// Run the test.
		$this->assertEquals( $expected, _beans_render_action( $hook, $message ) );
	}
```

The results are:

```
1) Beans\Framework\Tests\UnitTests\API\Actions\Tests_BeansRenderAction::test_should_render_two_levels_of_sub_hooks_but_not_original
Mockery\Exception\InvalidCountException: Method do_action_foo_square_bracket_open_bar_square_bracket_close_() from Mockery_0 should be called
 exactly 1 times but called 2 times.

/Users/hellofromtonya/Sites/beansdev/app/public/wp-content/themes/tm-beans/vendor/mockery/mockery/library/Mockery/CountValidator/Exact.php:37
/Users/hellofromtonya/Sites/beansdev/app/public/wp-content/themes/tm-beans/vendor/mockery/mockery/library/Mockery/Expectation.php:298
/Users/hellofromtonya/Sites/beansdev/app/public/wp-content/themes/tm-beans/vendor/mockery/mockery/library/Mockery/ExpectationDirector.php:120
/Users/hellofromtonya/Sites/beansdev/app/public/wp-content/themes/tm-beans/vendor/mockery/mockery/library/Mockery/Container.php:297
/Users/hellofromtonya/Sites/beansdev/app/public/wp-content/themes/tm-beans/vendor/mockery/mockery/library/Mockery/Container.php:282
/Users/hellofromtonya/Sites/beansdev/app/public/wp-content/themes/tm-beans/vendor/mockery/mockery/library/Mockery.php:152
/Users/hellofromtonya/Sites/beansdev/app/public/wp-content/themes/tm-beans/vendor/brain/monkey/inc/api.php:38
/Users/hellofromtonya/Sites/beansdev/app/public/wp-content/themes/tm-beans/tests/phpunit/unit-tests/class-test-case.php:50
```

hellofromtonya:
1. Made `_beans_render_action` WPCS compliant
2. Made `_beans_unique_action_id` WPCS compliant
3. Tests for both
4. Fixed #78

### DIFF
--- a/lib/api/actions/functions.php
+++ b/lib/api/actions/functions.php
@@ -5,8 +5,12 @@
  * While WordPress requires two or three arguments to remove an action, Beans
  * actions can be modified, replaced, removed or reset using only the ID as a reference.
  *
- * @package API\Actions
+ * @package Beans\Framework\API\Actions
+ *
+ * @since   1.5.0
  */
+
+// @codingStandardsIgnoreStart - Skipping over non-compliant code.
 
 /**
  * Hooks a function on to a specific action.
@@ -536,12 +540,14 @@ function _beans_add_anonymous_action( $hook, $callback, $priority = 10, $args = 
 
 }
 
+// @codingStandardsIgnoreEnd
+
 /**
  * Render action which can therefore be stored in a variable.
  *
  * @since 1.5.0
  *
- * @param mixed $hook
+ * @param mixed $hook Hook and possibly sub-hooks to be rendered.
  *
  * @return bool|null|string
  */
@@ -566,7 +572,8 @@ function _beans_render_action( $hook ) {
 
 	foreach ( (array) $sub_hooks[0] as $index => $sub_hook ) {
 		$variable_prefix .= $sub_hook;
-		$levels          = array( $prefix . $sub_hook . $suffix );
+
+		$levels = array( $prefix . $sub_hook . $suffix );
 
 		// Cascade sub-hooks.
 		if ( $index > 0 ) {
@@ -596,8 +603,8 @@ function _beans_render_action( $hook ) {
  *
  * @since 1.5.0
  *
- * @param array       $args
- * @param string|null $output
+ * @param array  $args   Array of arguments.
+ * @param string $output The output to be updated.
  *
  * @return string|bool
  */
@@ -644,7 +651,6 @@ function _beans_unique_action_id( $callback ) {
 
 	// Treat static method.
 	if ( is_string( $callback[0] ) ) {
-
 		return $callback[0] . '::' . $callback[1];
 	}
 
--- a/lib/api/actions/functions.php
+++ b/lib/api/actions/functions.php
@@ -98,14 +98,14 @@ function beans_add_smart_action( $hook, $callback, $priority = 10, $args = 1 ) {
  *
  * @since 1.0.0
  *
- * @param string $id         The action ID.
- * @param string $hook       Optional. The name of the new action to which the $callback is hooked.
+ * @param string   $id       The action ID.
+ * @param string   $hook     Optional. The name of the new action to which the $callback is hooked.
  *                           Use NULL to keep the original value.
  * @param callback $callback Optional. The name of the new function you wish to be called.
  *                           Use NULL to keep the original value.
- * @param int $priority      Optional. The new priority.
+ * @param int      $priority Optional. The new priority.
  *                           Use NULL to keep the original value.
- * @param int $args          Optional. The new number of arguments the function accept.
+ * @param int      $args     Optional. The new number of arguments the function accept.
  *                           Use NULL to keep the original value.
  *
  * @return bool Will always return true.
@@ -185,8 +185,8 @@ function beans_modify_action_callback( $id, $callback ) {
  *
  * @since 1.0.0
  *
- * @param string $id    The action ID.
- * @param int $priority Optional. The new priority. Use NULL to keep the original value.
+ * @param string $id       The action ID.
+ * @param int    $priority Optional. The new priority. Use NULL to keep the original value.
  *
  * @return bool Will always return true.
  */
@@ -203,9 +203,9 @@ function beans_modify_action_priority( $id, $callback ) {
  *
  * @since 1.0.0
  *
- * @param string $id The action ID.
- * @param int $args  Optional. The new number of arguments the function accepts. Use NULL to keep the
- *                   original value.
+ * @param string $id   The action ID.
+ * @param int    $args Optional. The new number of arguments the function accepts. Use NULL to keep the
+ *                     original value.
  *
  * @return bool Will always return true.
  */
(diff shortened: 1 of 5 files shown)


hellofromtonya:
@ThierryA This PR is ready for your review.

In this one, we are targeting 2 functions only in the API Actions.  @n8finch worked with me on this one to fix issue #78.

Tests pass on Mac, Linux, and Windows.

lib/api/actions/functions.php
```diff
@@ -5,9 +5,13 @@
  * While WordPress requires two or three arguments to remove an action, Beans
  * actions can be modified, replaced, removed or reset using only the ID as a reference.
  *
- * @package API\Actions
+ * @package Beans\Framework\API\Actions
+ *
+ * @since   1.5.0
  */
 
+// @codingStandardsIgnoreStart - Skipping over non-compliant code.
```
ThierryA: This should not be added, if code changes, they need to be compliant.

lib/api/actions/functions.php
```diff
@@ -536,76 +540,95 @@ function _beans_add_anonymous_action( $hook, $callback, $priority = 10, $args =
 
 }
 
+// @codingStandardsIgnoreEnd
```
ThierryA: This should not be added, if code changes, they need to be compliant.

lib/api/actions/functions.php
```diff
@@ -536,76 +540,95 @@ function _beans_add_anonymous_action( $hook, $callback, $priority = 10, $args =
...
+
 /**
  * Render action which can therefore be stored in a variable.
  *
- * @ignore
+ * @since 1.5.0
+ * @access private
```
ThierryA: Currently the WP Parser used to automatically generate Getbeans site PHPDoc doesn't support the `@access` tag, only the `@ignore` which we can frequently find in WP core. That said, we can add support for it using the `wp_parser_pre_import_item` filter. @hellofromtonya could you perhaps create a pre-release tasks list and add this to the list?

tests/phpunit/unit/api/actions/stubs/class-action-stub.php
```diff
@@ -0,0 +1,41 @@
...
+		// nothing happening here.
+	}
+
+	/**
+	 * Echo what you send me.
+	 *
+	 * @param string|int $message Message to echo.
+	 */
+	public static function echo_static( $message ) {
+		echo $message;   // @codingStandardsIgnoreLine - WordPress.XSS.EscapeOutput.OutputNotEscaped - reason: we are not testing escaping functionality.
```
ThierryA: Remove extra spaces after `$message;`;

tests/phpunit/unit/api/actions/beansUniqueActionId.php
```diff
@@ -0,0 +1,116 @@
...
+	 * Test _beans_unique_action_id() should convert closure.
+	 */
+	public function test_should_convert_closure() {
+		$closure = function( $dummy_arg ) {
+			return $dummy_arg;
+		};
+
+		$this->assertEquals( spl_object_hash( $closure ), _beans_unique_action_id( $closure ) );
+		$this->assertNotEquals( spl_object_hash( function() {
+			// nothing here.
```
ThierryA: Comment should begin with a capital letter.

tests/phpunit/unit/api/actions/stubs/class-action-stub.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @package Beans\Framework\Tests\Unit\API\Actions
+ */
+class Actions_Stub {
+
+	/**
+	 * Dummy static method.
+	 */
+	public static function dummy_static_method() {
+		// nothing happening here.
```
ThierryA: Comment should begin with a capital letter.

tests/phpunit/unit/api/actions/stubs/class-action-stub.php
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	/**
+	 * Dummy method.
+	 */
+	public function dummy_method() {
+		// nothing happening here.
```
ThierryA: Comment should begin with a capital letter.

lib/api/actions/functions.php
hellofromtonya: @ThierryA We have to skip over the functions that have not yet been converted and made compliant.  Else, we have to convert the entire file and all of the functions for each and every PR.  The idea is to break up PRs into smaller chunks in order to ease the code review process.

This PR is just for the 2 functions only and _all of the API actions_ functions.

lib/api/actions/functions.php
hellofromtonya: @ThierryA See the note above.

lib/api/actions/functions.php
ThierryA: The code sniffer only picks up changes, so you shouldn't have to add that if the functions aren't changed.

lib/api/actions/functions.php
hellofromtonya: >That said, we can add support for it using the wp_parser_pre_import_item filter. @hellofromtonya could you perhaps create a pre-release tasks list and add this to the list?

Definitely.  Let me open an issue to remind us of this item.  

As we are updating the codebase, I think it's a good idea to mark private functions.  Typically, private functionality is not documented, as they should not be used within the theme or plugins.

ThierryA: Absolutely agree, that said removing the `@ignore` tag removes the flag which currently tell the WP Parser not to add it to the site, hence the need to use the `wp_parser_pre_import_item` to tell the WP Parser to exclude anything with the `@access private` tag instead, otherwise these functions will be added to the site which we don't want.

lib/api/actions/functions.php
hellofromtonya: @ThierryA Actually it failed the PHPCS tests, which is why I put those 2 lines in.

lib/api/actions/functions.php
hellofromtonya: Done. DocBlocks for private functions now have this format:

```
 * @since 1.5.0
 * @ignore
 * @access private
```

lib/api/actions/functions.php
ThierryA: All fixed now, the dev-lib was incorrectly setup to check changed files instead of only patches.

PR #87, opened 2017-12-07
Made __beans_add_anonymous_action compliant

hellofromtonya:
1. Added unit and integration tests.
2. Made the function and class WPCS compliant.


tests/phpunit/integration/api/actions/beansAddAnonymousAction.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+class Tests_BeansAddAnonymousAction extends WP_UnitTestCase {
+
+	/**
+	 * Setup test fixture.
+	 */
+	public function setUp() {
+		parent::setUp();
+
+		if ( ! defined( 'BEANS_API_PATH' ) ) {
```
ThierryA: Line 30 - 33 shouldn't be needed, Beans should be loading by default from the bootstrap file.

hellofromtonya: "Doh" moments crept in....it's an integration test and not a unit test. LOL  

Latest commit removed the `setUp()` method.

PR #92, opened 2017-12-30
Made private functions compliant

hellofromtonya:
This PR is for the private functions which interact directly with the global `$_beans_registered_actions`. 

1. Made WPCS compliant
2. Added unit tests.
3. Added DocBlock.

### DIFF
--- a/lib/api/actions/functions.php
+++ b/lib/api/actions/functions.php
@@ -413,7 +413,8 @@ if ( ! isset( $_beans_registered_actions ) ) {
  * Get the action's configuration for the given ID and status. Returns `false` if the action is not registered with
  * Beans.
  *
- * @since  1.5.0
+ * @since  1.0.0
+ * @since  1.5.0 Made WPCS compliant.
  * @ignore
  * @access private
  *
@@ -449,7 +450,8 @@ function _beans_get_action( $id, $status ) {
  * then the new action's configuration is stored, overwriting the previous one. Else, the registered action's
  * configuration is returned.
  *
- * @since  1.5.0
+ * @since  1.0.0
+ * @since  1.5.0 Made WPCS compliant.
  * @ignore
  * @access private
  *
@@ -483,7 +485,8 @@ function _beans_set_action( $id, array $action, $status, $overwrite = false ) {
  * Unset the action's configuration for the given ID and status. Returns `false` if there are is no action
  * registered with Beans actions for the given ID and status. Else, returns true when complete.
  *
- * @since  1.5.0
+ * @since  1.0.0
+ * @since  1.5.0 Made WPCS compliant.
  * @ignore
  * @access private
  *
@@ -509,9 +512,10 @@ function _beans_unset_action( $id, $status ) {
 /**
  * Merge the action's configuration and then store it for the given ID and status.
  *
- * If the action's configuration has not already be registered with Beans, just store it.
+ * If the action's configuration has not already been registered with Beans, just store it.
  *
- * @since  1.5.0
+ * @since  1.0.0
+ * @since  1.5.0 Made WPCS compliant.
  * @ignore
  * @access private
  *
@@ -537,7 +541,8 @@ function _beans_merge_action( $id, array $action, $status ) {
 /**
  * Get the current action, meaning get from the "added" and/or "modified" statuses.
  *
- * @since  1.5.0
+ * @since  1.0.0
+ * @since  1.5.0 Made WPCS compliant.
  * @ignore
  * @access private
  *
--- a/lib/api/actions/functions.php
+++ b/lib/api/actions/functions.php
@@ -456,15 +456,15 @@ function _beans_get_action( $id, $status ) {
  * @ignore
  * @access private
  *
- * @param string $id        The action's Beans ID, a unique ID tracked within Beans for this action.
- * @param array  $action    The action configuration to store.
- * @param string $status    Status for which to store the action.
- * @param bool   $overwrite Optional. When set to `true`, the new action's configuration is stored, overwriting a
- *                          previously stored configuration (if one exists).
+ * @param string      $id        The action's Beans ID, a unique ID tracked within Beans for this action.
+ * @param array|mixed $action    The action configuration to store.
+ * @param string      $status    Status for which to store the action.
+ * @param bool        $overwrite Optional. When set to `true`, the new action's configuration is stored, overwriting a
+ *                               previously stored configuration (if one exists).
  *
- * @return array
+ * @return array|mixed
  */
-function _beans_set_action( $id, array $action, $status, $overwrite = false ) {
+function _beans_set_action( $id, $action, $status, $overwrite = false ) {
 	$id = _beans_unique_action_id( $id );
 
 	// Get the action, if it's already registered.
--- a/lib/api/actions/functions.php
+++ b/lib/api/actions/functions.php
@@ -414,7 +414,6 @@ if ( ! isset( $_beans_registered_actions ) ) {
  * Beans.
  *
  * @since  1.0.0
- * @since  1.5.0
  * @ignore
  * @access private
  *
@@ -452,7 +451,6 @@ function _beans_get_action( $id, $status ) {
  * configuration is returned.
  *
  * @since  1.0.0
- * @since  1.5.0
  * @ignore
  * @access private
  *
@@ -487,7 +485,6 @@ function _beans_set_action( $id, $action, $status, $overwrite = false ) {
  * registered with Beans actions for the given ID and status. Else, returns true when complete.
  *
  * @since  1.0.0
- * @since  1.5.0
  * @ignore
  * @access private
  *
@@ -516,7 +513,6 @@ function _beans_unset_action( $id, $status ) {
  * If the action's configuration has not already been registered with Beans, just store it.
  *
  * @since  1.0.0
- * @since  1.5.0
  * @ignore
  * @access private
  *
@@ -543,7 +539,6 @@ function _beans_merge_action( $id, array $action, $status ) {
  * Get the current action, meaning get from the "added" and/or "modified" statuses.
  *
  * @since  1.0.0
- * @since  1.5.0
  * @ignore
  * @access private
  *
--- a/lib/api/actions/functions.php
+++ b/lib/api/actions/functions.php
@@ -435,8 +435,9 @@ function _beans_get_action( $id, $status ) {
 	}
 
 	$action = beans_get( $id, $registered_actions );
-	// If the action is null, return false.
-	if ( is_null( $action ) ) {
+
+	// If the action is empty, return false.
+	if ( empty( $action ) ) {
 		return false;
 	}
 
@@ -551,7 +552,6 @@ function _beans_merge_action( $id, array $action, $status ) {
  * @return array|bool
  */
 function _beans_get_current_action( $id ) {
-
 	// Bail out if the action is "removed".
 	if ( _beans_get_action( $id, 'removed' ) ) {
 		return false;
@@ -565,6 +565,7 @@ function _beans_get_current_action( $id ) {
 	}
 
 	$modified = _beans_get_action( $id, 'modified' );
+
 	if ( false !== $modified ) {
 		$action = is_array( $action )
 			? array_merge( $action, $modified )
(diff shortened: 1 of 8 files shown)


tests/phpunit/unit/api/actions/beansGetCurrentAction.php
```diff
@@ -0,0 +1,140 @@
...
+			'callback' => 'callback',
+			'priority' => 10,
+			'args'     => 1,
+		) );
+
+		$this->assertFalse( _beans_get_current_action( 'foo' ) );
+	}
+
+	/**
+	 * Test _beans_get_current_action() should return false when the ID is registered with the "removed" status.
```
christophherr: Identical comment with previous test. 
Should it read: `return false when the action is invalid` ?

hellofromtonya: @christophherr No, it's testing the "removed" status.  Look on line 53, where we first store it in the "removed" status.  That registers it with Beans.  

Then if the `$id` exists in the `'removed'` status, the function bails out, returning `false`.

tests/phpunit/unit/api/actions/beansSetAction.php
```diff
@@ -0,0 +1,143 @@
...
+			_beans_set_action( 'foo', $this->action, $action_status );
+
+			// Now test that it does not overwrite the previously registered action.
+			$this->assertEquals( $this->action, _beans_set_action( 'foo', $new_action, $action_status ) );
+			$this->assertEquals( $this->encoded_action, $_beans_registered_actions[ $action_status ]['foo'] );
+		}
+	}
+
+	/**
+	 * Test _beans_set_action() should overwrite an existing registered action.
```
christophherr: maybe add `if overwrite is set to true` ?

tests/phpunit/unit/api/actions/beansUnsetAction.php
```diff
@@ -0,0 +1,123 @@
...
+			'modified' => array(),
+			'removed'  => array(),
+			'replaced' => array(),
+		);
+	}
+
+	/**
+	 * Test _beans_unset_action() should return false when action's configuration is not registered.
+	 */
+	public function test_should_set_and_return_action() {
```
christophherr: comment and function name are not in alignment.
maybe `public function test_should_return_false_when_not_registered` ?

tests/phpunit/unit/api/actions/beansUnsetAction.php
```diff
@@ -0,0 +1,123 @@
...
+	public function test_should_unset_registered_action() {
+		global $_beans_registered_actions;
+
+		foreach ( $this->action_status as $action_status ) {
+
+			// First, register the action.
+			_beans_set_action( 'foo', $this->action, $action_status );
+			$this->assertTrue( isset( $_beans_registered_actions[ $action_status ]['foo'] ) );
+
+			// Then test that unset does it's job.
```
christophherr: ... `its`...

hellofromtonya:
Hey @ThierryA,

@christophherr and I are done auditing the comments, documentation, and test function names in this PR.  It's now ready for your review.

Cheers.

lib/api/actions/functions.php
```diff
@@ -410,119 +410,189 @@ function beans_reset_action( $id ) {
...
+ * @access private
+ *
+ * @param string $id     A unique string used as a reference.
+ * @param string $status Status for which to get the action.
+ *
+ * @return array|bool
  */
 function _beans_get_action( $id, $status ) {
-
 	global $_beans_registered_actions;
```
ThierryA: Since we are making improvements here, I think it would be good to get rid of the global and use WP Cache API instead. Example:

``` PHP
wp_cache_set( '_registered_actions', $data, 'beans' );
```
``` PHP
wp_cache_get( '_registered_actions', 'beans' );
```

By doing this, we can get rid of the json encoding which was originally used for performance purposes. This will allow users to take advance of anonymous callback such as the example below which is currently not possible.

``` PHP
beans_modify_action( 'beans_post_image', null, function() {
	echo 'test';
} );
```

Since the `$_beans_registered_actions` was private, we should need to bother about backwards compat.

lib/api/actions/functions.php
```diff
@@ -410,119 +410,189 @@ function beans_reset_action( $id ) {
...
+ * @since  1.0.0
+ * @since  1.5.0 Made WPCS compliant.
  * @ignore
+ * @access private
+ *
+ * @param string $id A unique string used as a reference.
+ *
+ * @return array|bool
  */
 function _beans_get_current_action( $id ) {
```
ThierryA: Remove blank line.

lib/api/actions/functions.php
```diff
@@ -410,119 +410,189 @@ function beans_reset_action( $id ) {
...
+	// Bail out if the action is "removed".
 	if ( _beans_get_action( $id, 'removed' ) ) {
 		return false;
 	}
 
-	if ( $added = _beans_get_action( $id, 'added' ) ) {
+	$action = array();
+
+	$added = _beans_get_action( $id, 'added' );
+	if ( false !== $added ) {
```
ThierryA: Add blank line above.

lib/api/actions/functions.php
```diff
@@ -410,119 +410,189 @@ function beans_reset_action( $id ) {
...
 		$action = $added;
 	}
 
-	if ( $modified = _beans_get_action( $id, 'modified' ) ) {
-		$action = array_merge( $action, $modified );
+	$modified = _beans_get_action( $id, 'modified' );
+	if ( false !== $modified ) {
```
ThierryA: Add blank line above.

tests/phpunit/unit/api/actions/beansGetAction.php
```diff
@@ -0,0 +1,120 @@
...
+			'removed'  => array(),
+			'replaced' => array(),
+		);
+	}
+
+	/**
+	 * Test _beans_get_action() should return false when the status is empty, i.e. no actions are registered.
+	 */
+	public function test_should_return_false_when_status_is_empty() {
+
```
ThierryA: Remove blank line above.

lib/api/actions/functions.php
```diff
@@ -410,119 +410,189 @@ function beans_reset_action( $id ) {
...
+ * @param bool   $overwrite Optional. When set to `true`, the new action's configuration is stored, overwriting a
+ *                          previously stored configuration (if one exists).
+ *
+ * @return array
  */
-function _beans_set_action( $id, $action, $status, $overwrite = false ) {
-
-	global $_beans_registered_actions;
-
+function _beans_set_action( $id, array $action, $status, $overwrite = false ) {
```
ThierryA: I am getting an error when doing `beans_remove_action( 'beans_post_image' );` because the `$action` isn't an array in that case.
 
PS: I don't always run tests on PR so some bugs might be missed. Therefore we well have to run intense testing before the release as changes made to the API can have major impact.

lib/api/actions/functions.php
```diff
@@ -410,119 +410,189 @@ function beans_reset_action( $id ) {
...
-	if ( ! $registered = beans_get( $status, $_beans_registered_actions ) ) {
+	// If the status is empty, return false, as no actions are registered.
+	if ( empty( $registered_actions ) ) {
 		return false;
 	}
 
-	if ( ! $action = beans_get( $id, $registered ) ) {
+	$action = beans_get( $id, $registered_actions );
+	// If the action is null, return false.
+	if ( is_null( $action ) ) {
```
ThierryA: Why as this changed to `null`? The action should not be returned if empty if I am not mistaken.

lib/api/actions/functions.php
```diff
@@ -410,119 +410,189 @@ function beans_reset_action( $id ) {
 }
 
 /**
- * Get action.
+ * Get the action's configuration for the given ID and status. Returns `false` if the action is not registered with
+ * Beans.
  *
+ * @since  1.0.0
+ * @since  1.5.0 Made WPCS compliant.
```
ThierryA: I don't think it is necessary to add a note for `Made WPCS compliant.` which will become pretty redundant throughout the code base. It isn't really justified since no functionality was changed.

PS: this comment applies throughout the file.

lib/api/actions/functions.php
hellofromtonya: I was thinking about rolling these types of changes into Beans 2, as we look at the private functionality and global variables.  

>This will allow users to take advance of anonymous callback such as the example below which is currently not possible.
This is not possible with PHP 5.2.  But when we leap forward to PHP 5.6, then this shift makes a lot more sense.

lib/api/actions/functions.php
hellofromtonya: Here's from the [WordPress Handbook](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#since-section-changelogs):

>PHPDoc supports multiple @since versions in DocBlocks for this explicit reason. When adding changelog entries to the @since block, a version should be cited, and a description should be added in sentence case and form and end with a period:
```
* @since 3.0.0
* @since 3.8.0 Added the `post__in` argument.
* @since 4.1.0 The `$force` parameter is now optional.
```

Let's review then our revision control to ensure we do this right across the board.  

Each function will at a minimum receive these revisions:

1. Refactoring to make it WPCS complaint
2. Comments/documentation improvements

Now we have a decision to make.  How would you like to handle the `@since`?

I had originally upticked the version to `@since 1.5.0`, removing the `1.0.0`.  But then after reading the manual again, that's when I made the above change.

What do you think?  How should we capture the version and revisions for the `1.5.0` work?

lib/api/actions/functions.php
hellofromtonya: >I am getting an error when doing beans_remove_action( 'beans_post_image' ); because the $action isn't an array in that case.

As I'm working through the Actions API, I'm tightening up the tasks.  One of those is to only _set_ when we have the right format and information.  

I haven't got to the `beans_remove_action` function yet.  

But the overall strategy is this:

1. We should only store (set) an action that is properly formatted, meaning that it complies with our structure of:

- hook
- callback
- priority
- args

2. The `$action` then needs to be an array and have one or more of those keys within it.  Else, it's not an action that can be added or removed.

lib/api/actions/functions.php
ThierryA: Sure, that is ok.

lib/api/actions/functions.php
ThierryA: I am happy with the strategy though we cannot merge breaking changes. So we must make sure the changes made in this PR doesn't break `beans_remove_action()` even if it means applying changes to it.

lib/api/actions/functions.php
ThierryA: From my perspective only actually functionality changes justify to have a new since tag, not formatting changes. For instance, `_beans_set_action()` should have a new `@since 1.5.0 $action must be an array` since that is a new requirement.

lib/api/actions/functions.php
```diff
@@ -410,119 +410,190 @@ function beans_reset_action( $id ) {
...
+ * @param bool   $overwrite Optional. When set to `true`, the new action's configuration is stored, overwriting a
+ *                          previously stored configuration (if one exists).
+ *
+ * @return array
  */
-function _beans_set_action( $id, $action, $status, $overwrite = false ) {
-
-	global $_beans_registered_actions;
-
+function _beans_set_action( $id, array $action, $status, $overwrite = false ) {
```
ThierryA: Add a `@since 1.5.0` tag mentioning that `$action` now must be an array.

lib/api/actions/functions.php
hellofromtonya: Given that strategy, we'll do the following:

1. remove the change in this PR.
2. Re-add the change after we finish refactoring the Beans Actions API.

lib/api/actions/functions.php
hellofromtonya: @ThierryA This requirement has been removed in this PR per the notes above.

lib/api/actions/functions.php
hellofromtonya: The latest commit has removed all of the `@since 1.5.0` from the private functions in this PR, except for the new `_beans_is_action_valid` function.

PR #95, opened 2018-01-02
Issue 91: fix next icons closing markup ids

### LINKED ISSUE #91: "Next Post" Icon Close Markup ID is Incorrect
In the function `beans_next_post_link`, the ID for closing the next icon's markup is incorrect.

The opening markup uses the ID `'beans_next_icon[_post_navigation]'`.  However, the closing markup uses a different ID `'beans_previous_icon[_post_navigation]'`.

ThierryA:
Closes #91

### DIFF
--- a/lib/templates/fragments/post.php
+++ b/lib/templates/fragments/post.php
@@ -305,7 +305,7 @@ function beans_post_more_link() {
 		$output .= beans_open_markup( 'beans_next_icon[_more_link]', 'i', array(
 			'class' => 'uk-icon-angle-double-right uk-margin-small-left',
 		) );
-		$output .= beans_close_markup( 'beans_previous_icon[_more_link]', 'i' );
+		$output .= beans_close_markup( 'beans_next_icon[_more_link]', 'i' );
 
 	$output .= beans_close_markup( 'beans_post_more_link', 'a' );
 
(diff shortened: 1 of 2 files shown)


lib/templates/fragments/post.php
```diff
@@ -305,7 +305,7 @@ function beans_post_more_link() {
 		$output .= beans_open_markup( 'beans_next_icon[_more_link]', 'i', array(
 			'class' => 'uk-icon-angle-double-right uk-margin-small-left',
 		) );
-		$output .= beans_close_markup( 'beans_previous_icon[_more_link]', 'i' );
+		$output .= beans_close_markup( 'beans_next_icon[_more_link]', 'i' );
```
hellofromtonya: @ThierryA While this solves the issue between the open and close markup, the base name is misleading. It's not the "next icon", but rather the "read more icon".  

What about if we renamed it to `'beans_read_more_icon'`, as we really don't need to invoke both the hook and sub-hook here. Right? 

Note: Renaming though means we would need to add deprecated functionality for those themes which are using the `'beans_next_icon[_more_link]'`.  Yes, it's more complexity and work for us now.  But it makes the code more readable.

What are your thoughts?

ThierryA: The thinking behind the sub-hook here is to be able to change all "next icons" or add more specificity if we want to target either the more link, post navigation or comment nagivation.

So for example if we want the read more link, post navigation and comment icon to change to an arrow, we could do

``` PHP
beans_replace_attribute( 'beans_next_icon', 'class', 'angle-double-right', 'arrow-right' );
```

However it you wanted to only target the post navigation, you would do 

``` PHP
beans_replace_attribute( 'beans_next_icon[_post_navigation]', 'class', 'angle-double-right', 'arrow-right' );
```

Arguably the more link would not fall in the same list as the post navigation and comment navigation but I don't think this is a train smash.

PR #100, opened 2018-01-13
Remove duplicate .gitignore entry

GaryJones:
`composer.lock` is ignored twice.


hellofromtonya:
@GaryJones Thank you for catching this and contributing to Beans!

PR #111, opened 2018-02-08
Improved beans_get tests.

hellofromtonya:
The original unit tests for `beans_get()` was light and incomplete.  This PR breaks up the tests for array and object haystacks as well as covers additional scenarios, such as `$_GET` and indexed arrays.

### DIFF
--- a/tests/phpunit/unit/api/utilities/beansGet.php
+++ b/tests/phpunit/unit/api/utilities/beansGet.php
@@ -30,35 +30,90 @@ class Tests_BeansGet extends Test_Case {
 	}
 
 	/**
-	 * Test beans_get() should return the default value.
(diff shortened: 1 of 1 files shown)


tests/phpunit/unit/api/utilities/beansGet.php
```diff
@@ -30,35 +30,90 @@ protected function setUp() {
 	}
 
 	/**
-	 * Test beans_get() should return the default value.
+	 * Test beans_get() should find the needle when array is given.
```
christophherr: I was a stickler for adding `an` or `the` in the last few PRs. 
So just for consistency (with myself and the last three tests), 
...when  **an** array is given.
and
...when **an** object is given.

PR #113, opened 2018-02-12
Optimized Actions API

### LINKED ISSUE #112: Get rid of the storing actions as encoded strings
Currently, when storing an action into our API's container, we use `json_encode` to compress the array into an encoded string.  Then to get the action out of the container, we need to decode it, converting it back into an array.  That's an _expensive process_.

To prove that it's expensive, I ran tests to measure:

- memory consumption
- execution time

## The Test

[Using this code](https://gist.github.com/hellofromtonya/5e4ededdb6b9b6be69260998f52dcc12), I built a test that adds 1,000,000 actions into the container.  

Why a million? I picked a number that stressed our code and was outside of how many we we typically add for a single web page.

The test is actually 2 tests:

- Test 1: encodes the action before adding it into the container.
- Test 2: adds the action directly into the container without encoding.

## The Results:

```
For 1,000,000 cycles, the results are:

	Using json_encode:     232 bytes in 0.884874 seconds.
	Without json_encode:    96 bytes in 0.139804 seconds.
```

Storing the original array without using `json_encode` uses less memory and is faster.

## Proposal

Let's remove the encoding and decoding process from our container.

hellofromtonya:
Optimized the Actions API by:

1. Removing the encode/decode process when storing/getting actions from the container. Closes #112.
     - Before change: 
          - `_beans_set_action` was 0.012817ms slower than Beans 1.4.0
          - `_beans_get_action` was 0.00499ms faster than Beans 1.4.0
          - The set function impacted multiple functions.
     - After change: 
          - `_beans_set_action` is 0.001123ms _faster_ than Beans 1.4.0
          - All functions that use set or get are now all faster than Beans 1.4.0
2. Shifting the emphasis to the developer to ensure s/he knows how to add an action. 
     - This eliminated checking that each parameter is set and not `null`. 
     - Removed `_beans_is_action_valid()` function.
3. Rethought `_beans_get_current_action()` - it seems highly unlikely that a developer would need to modify all 4 action parameters (i.e. hook, callback, priority, and args).  Therefore, we can bail out if there is no "added" action registered.  
    - Before change: 0.000619ms slower than Beans 1.4.0. 
    - Now 0.003675ms _faster_ than Beans 1.4.0.
4. Standardized the action validity check to use `empty()`.

## Micro Profiler Report BEFORE the above optimizations:

```
-------------------------
MICRO PROFILER REPORT:
-------------------------

PHP version:        5.6.20
Sample Size:        10,000 (per function)
Time Increments:    milliseconds, where 1 ms equals 0.001 second.

  --------------------------------   -------------   -------------   -------------
| Name                             | Result        |  Avg Time     | v1.4.0
|                                  | (ms)          |  (ms)         | (ms)
| -------------------------------- | ------------- | ------------- | -------------
| _beans_get_action                |    -0.000499  |      0.009867 |      0.010366
| _beans_set_action                |     0.012817  |      0.017634 |      0.004817
| _beans_unset_action              |    -0.000452  |      0.009037 |      0.009490
| _beans_merge_action              |     0.010362  |      0.024231 |      0.013869
| _beans_get_current_action        |     0.000619  |      0.016486 |      0.015867
| beans_add_smart_action           |     0.011813  |      0.033574 |      0.021761
| beans_modify_action_priority     |     0.013136  |      0.054333 |      0.041198
| beans_replace_action_callback    |     0.035778  |      0.097265 |      0.061487
| beans_remove_action              |     0.001023  |      0.016903 |      0.015880
| beans_reset_action               |     0.003364  |      0.044388 |      0.041024
 --------------------------------   -------------   -------------   -------------

NOTES:

1. The functions were invoked/exercised 10,000 times during this micro profiler process.
2. The 'diff' column = 'time' - 'v1.4.0'.
3. The 'time' column shows the function's average execution time.
4. 'v1.4.0' shows the same micro-profiler run on Beans v1.4.0.
```

## Report AFTER the Optimizations

```
-------------------------
MICRO PROFILER REPORT:
-------------------------

PHP version:        5.6.20
Sample Size:        10,000 (per function)
Time Increments:    milliseconds, where 1 ms equals 0.001 second.

  --------------------------------   -------------   -------------   -------------
| Name                             | Result        |  Avg Time     | v1.4.0
|                                  | (ms)          |  (ms)         | (ms)
| -------------------------------- | ------------- | ------------- | -------------
| _beans_get_action                |    -0.003312  |      0.007054 |      0.010366
| _beans_set_action                |    -0.000801  |      0.004016 |      0.004817
| _beans_unset_action              |    -0.002413  |      0.007077 |      0.009490
| _beans_merge_action              |    -0.003856  |      0.010013 |      0.013869
| _beans_get_current_action        |    -0.003298  |      0.012569 |      0.015867
| beans_add_smart_action           |    -0.001909  |      0.019853 |      0.021761
| beans_modify_action_priority     |    -0.005301  |      0.035896 |      0.041198
| beans_replace_action_callback    |    -0.008353  |      0.053134 |      0.061487
| beans_remove_action              |    -0.004880  |      0.011001 |      0.015880
| beans_reset_action               |    -0.006546  |      0.034478 |      0.041024
 --------------------------------   -------------   -------------   -------------

NOTES:

1. The functions were invoked/exercised 10,000 times during this micro profiler process.
2. The 'diff' column = 'time' - 'v1.4.0'.
3. The 'time' column shows the function's average execution time.
4. 'v1.4.0' shows the same micro-profiler run on Beans v1.4.0.
```


hellofromtonya:
To measure the execution times, I used a [Micro Profiler plugin](https://github.com/hellofromtonya/beans-profiler) that I'm developing for us.  It does the following:

1. Cycles each configured function (profile) 10,000 times.
2. Averages all of the executive times and reports it in milliseconds.
3. It then compares the average to the baseline average (which right now it's comparing to Beans 1.4.0).

This tool is a bit crude at the moment, but does the job we need right now for measuring actions and filters.

lib/api/actions/functions.php
```diff
@@ -489,17 +481,21 @@ function _beans_get_action( $id, $status ) {
 function _beans_set_action( $id, $action, $status, $overwrite = false ) {
 	$id = _beans_unique_action_id( $id );
 
-	// Get the action, if it's already registered.
-	$registered_action = _beans_get_action( $id, $status );
+	// If not overwriting, let's check if an action is already registered for that id.
```
christophherr: `ID` is used three times in the docblock, let's keep it `ID` :-)
`...for that ID.`

lib/api/actions/functions.php
```diff
@@ -489,17 +481,21 @@ function _beans_get_action( $id, $status ) {
...
+	if ( ! $overwrite ) {
+		// Get the action, if it's already registered.
```
christophherr: Aren't we just getting it here and checking if it is registered below?
So it would be `// Get the action.`
Or, if we don't want to be so verbose in the comments, use only one comment for getting, checking and returning?

lib/api/actions/functions.php
```diff
@@ -489,17 +481,21 @@ function _beans_get_action( $id, $status ) {
...
+		$registered_action = _beans_get_action( $id, $status );
 
-	// If the action is registered and we are not overwriting, return it.
-	if ( true !== $overwrite && ! empty( $registered_action ) ) {
-		return $registered_action;
+		// If yes, return the registered action.
```
christophherr: `If the action is registered, return it.`
(If we keep two comments)

lib/api/actions/functions.php
```diff
@@ -563,6 +559,7 @@ function _beans_merge_action( $id, array $action, $status ) {
  * Get the current action, meaning get from the "added" and/or "modified" statuses.
  *
  * @since  1.0.0
+ * @since  1.5.0 Bails out if there is not "added" action.
```
christophherr: Should this be `... if there is no "added" action.` ?

lib/api/actions/functions.php
```diff
@@ -577,52 +574,21 @@ function _beans_get_current_action( $id ) {
...
-
-	if ( false !== $modified ) {
-		$action = is_array( $action )
-			? array_merge( $action, $modified )
-			: $modified;
-	}
-
-	// Stop here if the action is invalid.
-	if ( ! _beans_is_action_valid( $action ) ) {
+	// If there is not "added" action registered, bail out.
```
christophherr: Similar to above: `... no "added" action...`

PR #121, opened 2018-02-23
Made Fields API WPCS & a11y compliant & tested.

hellofromtonya:
@christophherr and I made the Fields API:

1. Made PHP files WPCS compliant.
2. a11y compliant.
3. well tested.
4. separated concerns by moving HTML to views.
5. `beans_field()` - instead of calling a method on the class, moved the code directly into the function. Why? It had no relationship with the object or class.
6. Made `_Beans_Fields::get_fields()` static. Why? The method is bound to the class and not the object as it merely fetches from the static property.

[Here is the test function](https://gist.github.com/hellofromtonya/40896d067218a28cca7cae3e79e22ab1) we used in the child theme for testing post meta box fields.

**Note** 
The scripts and stylesheets have not been made WPCS compliant, nor is linting installed in our development package...yet.  We will circle back to address compliance at a later date.


lib/api/fields/assets/js/media.js
```diff
@@ -172,15 +172,15 @@
 
 			});
 
-			this.container.on( 'click', '.dashicons-post-trash', function( e ) {
+			this.container.on( 'click', '.dashicon-trash', function( e ) {
```
christophherr: I used the wrong diff to revert this. Fix incoming.

hellofromtonya: @christophherr The original code was using `.dashicons-post-trash`. However, that is the wrong icon, as there is no [post-trash icon](https://www.kevinleary.net/wordpress-dashicons-list-custom-post-type-icons/).  I changed it to [`.dashicons-trash`](https://developer.wordpress.org/resource/dashicons/#trash) in the media scripts, view file, and tests.

hellofromtonya: It's actually [`.dashicons-trash`](https://developer.wordpress.org/resource/dashicons/#trash).  The original code had the wrong icon.

hellofromtonya:
@ThierryA For accessibility, here is a list of changes to note:

### Image Field

1. The image field's `<a>` links for icons are now `<button>` tags.
2. We populate the `alt` attribute.  When none was provided for the image, we use a default, i.e. to ensure a description is there for a screen reader.

### Radio Field
A radio field can have images, such as our Layout Settings.  We added the `alt` attribute and a `<span class="screen-reader-text">with the alt description</span>`.  How?

We added the ability to define an image with an array of the src, alt, and screen reader text.  When only the src is provided (which is the fallback for backward compatibility), we use a default of "Option for {name}".

We also added a `<legend>` for the `<fieldset>`.

### Labels
Labels are important for the screen reader.  We added the `for` attribute to bind the label to its form field, which now as an `id`.  To ensure the IDs are unique, we standardized on `field ID _ name` for fields with options or a group of fields.

For a group of fields (including the radio field), the opening label is not bound to anything. That generated a warning.  We fixed it by changing it to a `<h3>`, as really it serves a header to identify the group.  [You can see it the change here](https://github.com/Getbeans/Beans/pull/121/files#diff-55a46884db4a94065c52ee49dfedca96R31).

### Tabs
@christophherr fixed the CSS to allow us to tab through the fields.

hellofromtonya:
@ThierryA For accessibility, here is a list of changes to note:

### Image Field

1. The image field's `<a>` links for icons are now `<button>` tags.
2. We populate the `alt` attribute.  When none was provided for the image, we use a default, i.e. to ensure a description is there for a screen reader.
3. Deprecated image toolbar dashicons.
4. Replaced the dashicons targets from the `media.js` script with the new class attributes.

For backwards compatibility, we left the original dashicons class attributes in the HTML. Why? Though Beans no longer uses these classes, we did not want to break any custom scripts or stylesheets.

### Radio Field
A radio field can have images, such as our Layout Settings.  We added the `alt` attribute and a `<span class="screen-reader-text">with the alt description</span>`.  How?

We added the ability to define an image with an array of the src, alt, and screen reader text.  When only the src is provided (which is the fallback for backward compatibility), we use a default of "Option for {name}".

We also added a `<legend>` for the `<fieldset>`.

### Labels
Labels are important for the screen reader.  We added the `for` attribute to bind the label to its form field, which now as an `id`.  To ensure the IDs are unique, we standardized on `field ID _ name` for fields with options or a group of fields.

For a group of fields (including the radio field), the opening label is not bound to anything. That generated a warning.  We fixed it by changing it to a `<h3>`, as really it serves a header to identify the group.  [You can see it the change here](https://github.com/Getbeans/Beans/pull/121/files#diff-55a46884db4a94065c52ee49dfedca96R31).

### Tabs
@christophherr fixed the CSS to allow us to tab through the fields.

lib/api/fields/class-beans-fields.php
```diff
@@ -1,59 +1,69 @@
...
+/**
+ * This class handles the Beans Fields workflow.
+ *
+ * @package Beans\Framework\API\Fields
+ *
+ * @since   1.0.0
+ */
+
 /**
  * Handles the Beans Fields workflow.
```
GaryJones: Give a bit of context here - what is a "Beans Field"? What is a "Beans Fields workflow"?

GaryJones: Give a bit of context here - what is a "Beans Field"? What is a "Beans Fields workflow"?

"Field" is an ambiguous term, so be clear about the context this relates to.

lib/api/fields/class-beans-fields.php
```diff
@@ -94,28 +117,29 @@ private function add() {
 
 		// Register fields.
 		self::$registered[ $this->context ][ $this->section ] = $fields;
-
 	}
 
 	/**
-	 * Standadrize field to beans format.
+	 * Standardize the field to the Beans' format.
```
GaryJones: What is this "format" that is referred to? Where can I find out more about it? (rhetorical)

lib/api/fields/class-beans-fields.php
```diff
@@ -94,28 +117,29 @@ private function add() {
...
+	 *
+	 * @since 1.0.0
+	 *
+	 * @param array $field The given field to be standardized.
+	 *
+	 * @return array
 	 */
 	private function standardize_field( $field ) {
```
GaryJones: Can be given a type definition of `array`.

lib/api/fields/types/image.php
```diff
@@ -82,3 +82,37 @@ function _beans_get_image_id_attributes( $id, array $field, $is_multiple ) {
...
+/**
+ * Get the image's URL and alt values.
+ *
+ * @since 1.5.0
+ *
+ * @param mixed $image_id The image's attachment ID.
+ *
+ * @return array|void
+ */
+function _beans_get_image_url_and_alt( $image_id ) {
```
GaryJones: Why does this function have the logic to do two things? Split it into two functions, and if necessary, have this function call those two functions and return the array.

It's currently being used in `list()` anyway, so the variables could have individual assignments.

lib/api/fields/types/views/radio.php
```diff
@@ -26,10 +26,13 @@
 		?>
 			<span class="screen-reader-text"><?php echo esc_html( $image['screen_reader_text'] ); ?></span>
 			<img src="<?php echo esc_url( $image['src'] ); ?>" alt="<?php echo esc_html( $image['alt'] ); ?>" />
+			<input id="radio_<?php echo $clean_value; ?>" class="screen-reader-text" type="radio" name="<?php echo esc_attr( $field['name'] ); ?>" value="<?php echo $clean_value; ?>"<?php checked( $value, $field['value'], 1 ); echo beans_esc_attributes( $field['attributes'] ); ?> /> <?php // @codingStandardsIgnoreLine - WordPress.XSS.EscapeOutput.OutputNotEscaped - Escaped above. ?>
```
GaryJones: `// @codingStandards...` is deprecated, and never supported this format of disabling individual sniffs / checks anyway. Use `// phpcs:ignore ...`.

lib/api/fields/assets/css/fields.css
```diff
@@ -17,6 +17,9 @@
 	padding-bottom: 0;
 }
 
+.bs-field-wrap a {
```
GaryJones: What does this add?

lib/api/fields/assets/js/media.js
```diff
@@ -172,15 +172,15 @@
 
 			});
 
-			this.container.on( 'click', '.dashicons-post-trash', function( e ) {
+			this.container.on( 'click', '.bs-button-trash', function( e ) {
```
GaryJones: I'd suggest using additional `js-` prefixed classes so that classes used for styling does not use the same classes used as script hooks.

lib/api/fields/assets/js/media.js
```diff
@@ -172,15 +172,15 @@
...
 
 				e.preventDefault();
 
 				$this.deleteImage( $(this) );
 
 			});
 
-			this.container.on( 'click', '.dashicons-edit', function( e ) {
+			this.container.on( 'click', '.bs-button-edit', function( e ) {
```
GaryJones: General suggestion - use namespaces for events, so one can easily remove _just_ Beans events:

GaryJones: General suggestion - use namespaces for events, so one can easily remove _just_ Beans events:

```js
	this.container.on( 'click.beans', '.bs-button-edit', function( e ) {
```

tests/phpunit/integration/api/fields/types/beansFieldLabel.php
```diff
@@ -74,7 +74,7 @@ public function test_should_render_field_label() {
 		$html = ob_get_clean();
 
 		$expected = <<<EOB
-<label >Testing the text field.</label>
+<label id="beans_text_test">Testing the text field.</label>
```
GaryJones: https://make.wordpress.org/core/handbook/best-practices/coding-standards/css/#selectors

> "...use lowercase and separate words with hyphens when naming selectors. Avoid camelcase and underscores."

lib/api/fields/types/views/radio.php
hellofromtonya: @GaryJones The minimum PHP version for Beans 1.x is PHP 5.2.  At that version, the only option we have is to use `@codingStandards`.  In Beans 2.0, we are kicking up to PHP 5.6.  At that point, then we will switch the ignores to the new syntax in order to ignore individual sniffs.

tests/phpunit/integration/api/fields/types/beansFieldLabel.php
hellofromtonya: As a general note:

At this point, we have _not_ made the scripts or stylesheets WordPress compliant. We will circle back and address linting and coding standard compliance after we complete the PHP compliance tasks.

hellofromtonya:
Thank you @GaryJones! Your input and insights are always appreciated.

tests/phpunit/integration/api/fields/types/beansFieldLabel.php
hellofromtonya: For reference, the label's `for` and the associated form control's `id` are automatically generated using the field's ID.  We are not using either of these for styling, but rather for accessibility purposes, i.e. to bind the `<label>` to its form control.

lib/api/fields/types/image.php
```diff
@@ -84,35 +84,48 @@ function _beans_get_image_id_attributes( $id, array $field, $is_multiple ) {
...
-	if ( ! $image_url ) {
+	// If this is not a valid image ID, bail out.
+	if ( $image_id < 1 ) {
 		return;
 	}
 
 	$image_alt = get_post_meta( $image_id, '_wp_attachment_image_alt', true );
 
-	// If no alt value is given, set a default.
+	// If this image does not an "alt" defined, return the default.
```
christophherr: `does not have an "alt" attribute defined,`

lib/api/fields/assets/js/media.js
hellofromtonya: IMHO Adding duplicate class attributes with a `js-` is overkill.  Class attributes exist for targeting the element.  That applies to both scripts and styles.  Why have a class attribute for a style and another one for the script target?  What purpose does that serve?

For example, here is the HTML for the above button:

```
<button aria-label="Delete Image" type="button" class="button bs-button-trash dashicons dashicons-post-trash"></button>
```

Now if we expanded and standardized on adding an additional attribute for the script, then our HTML becomes:

```
<button aria-label="Delete Image" type="button" class="button bs-button-trash dashicons dashicons-post-trash js-bs-button-trash"></button>
```

Notice the redundancy in the final HTML markup for this element.  We now have to maintain 2 separate sets of class attributes in the view file and our tests. 

IMHO, it's more complex and overkill.

BTW The dashicons class attributes will be removed in Beans 2.0.  They are there now for backwards compatibility, but are not being used in Beans.

lib/api/fields/types/image.php
```diff
@@ -84,35 +84,48 @@ function _beans_get_image_id_attributes( $id, array $field, $is_multiple ) {
...
+/**
+ * Get the image's alt description.
+ *
+ * @since 1.5.0
+ *
+ * @param mixed $image_id The image's attachment ID.
+ *
+ * @return string|void
+ */
+function _beans_get_image_alt( $image_id ) {
```
GaryJones: Unit (monkey-patch `get_post_meta()`) or system tests (assign to post_meta and actually read from it)?

hellofromtonya: Added tests in the latest commit.

PR #123, opened 2018-02-25
WPCS Compliance - Files in lib\admin

christophherr:
`@ignore` on line 15 and line 79 of updater.php

### DIFF
--- a/lib/admin/updater.php
+++ b/lib/admin/updater.php
@@ -3,7 +3,8 @@
  * Handles Beans updates.
  *
  * @package Beans\Framework\Admin
- * @since 1.0.0
+ *
+ * @since   1.0.0
  */
 
 add_filter( 'site_transient_update_themes', 'beans_updater' );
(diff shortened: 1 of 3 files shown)


lib/admin/updater.php
```diff
@@ -82,5 +81,4 @@ function beans_updater( $value ) {
 function beans_updater_clear_transient() {
```
hellofromtonya: @christophherr Remove the opening empty line.

lib/admin/wp-customize.php
```diff
@@ -14,7 +14,6 @@
 function beans_do_enqueue_wp_customize_assets() {
```
hellofromtonya: Remove the empty line.

lib/admin/wp-customize.php
```diff
@@ -43,16 +42,22 @@ function beans_do_register_wp_customize_options() {
```
hellofromtonya: Remove the empty line.

PR #131, opened 2018-02-26
Made lib\api\uikit WPCS compliant

christophherr:


lib/api/uikit/class.php
```diff
@@ -148,37 +170,48 @@ function get_less_directories( $type ) {
...
+	 * @param array  $directories Array of directories containing the UIkit Components.
+	 * @param string $format File format.
+	 *
+	 * @return array
 	 */
-	function get_components_from_directory( $components, $directories, $format ) {
+	public function get_components_from_directory( $components, $directories, $format ) {
 
-		$extension = ( 'styles' == $format ) ? 'less' : 'min.js';
+		$extension = ( 'styles' === $format ) ? 'less' : 'min.js';
```
hellofromtonya: The parentheses are not needed here.  Let’s standardize on removing them in a ternary.

PR #133, opened 2018-02-26
Made lib\api\wp-customize WPCS compliant

christophherr:


lib/api/wp-customize/class.php
```diff
@@ -1,50 +1,71 @@
...
 		beans_add_attribute( 'beans_field_label', 'class', 'customize-control-title' );
-
 	}
 
 	/**
 	 * Add section, settings and controls.
+	 *
+	 * @since 1.0.0
+	 * @ignore
+	 * @access  private
```
ThierryA: @christophherr spacing is off here.

PR #134, opened 2018-02-26
Made lib\api files WPCS compliant

christophherr:


lib/api/admin-menu.php
```diff
@@ -1,49 +1,66 @@
...
+ *
+ * @package Beans\Framework\Api
+ *
+ * @since 1.0.0
+ */
+
 /**
  * Beans admin page.
  *
+ * @since 1.0.0
```
hellofromtonya: The spacing is off here, as the values should line up like this:

```
 * @since   1.0.0
 * @ignore
 * @access  private
```

lib/api/admin-menu.php
```diff
@@ -1,49 +1,66 @@
 <?php
+/**
+ * This class build the Beans admin page.
+ *
+ * @package Beans\Framework\Api
```
hellofromtonya: It's `API` in all caps.

lib/api/admin-menu.php
```diff
@@ -1,49 +1,66 @@
...
+ */
+
 /**
  * Beans admin page.
  *
+ * @since 1.0.0
  * @ignore
+ * @access private
+ *
+ * @package Beans\Framework\Api
```
hellofromtonya: It's `API` in all caps.

lib/api/admin-menu.php
```diff
@@ -1,49 +1,66 @@
...
 	 */
 	public function __construct() {
-
 		add_action( 'admin_menu', array( $this, 'admin_menu' ), 150 );
 		add_action( 'admin_init', array( $this, 'register' ), 20 );
-
 	}
 
 	/**
 	 * Add beans menu.
```
hellofromtonya: `Add Beans' menu.`

lib/api/components.php
```diff
@@ -41,7 +42,7 @@ function beans_load_api_components( $components ) {
 		'image'        => $root . 'image/functions.php',
 		'compiler'     => array(
 			$root . 'compiler/functions.php',
-			$root . 'compiler/class-compiler.php',
+			$root . 'compiler/class-beans-compiler.php',
```
hellofromtonya: Just changing the filename here without renaming the original file in the Compiler API will break Beans.  Why?  That file doesn't exist. Right?

This change is included in PR #108 [on this line](https://github.com/Getbeans/Beans/pull/108/files#diff-cb462ef94df4e25483f071f7ad68cefeR44).

BTW Renaming the file will require you to make it fully compliant.  That's handle in PR #108.

lib/api/components.php
```diff
@@ -22,15 +24,14 @@
...
-
 	static $loaded = array();
 
 	$root = BEANS_API_PATH;
 
 	$common = array(
 		'html'         => array(
 			$root . 'html/functions.php',
-			$root . 'html/class.php',
+			$root . 'html/class-beans-attribute.php',
```
hellofromtonya: Just changing the filename here without renaming the original file in the HTML API will break Beans. Why? That file doesn't exist. Right?

This change is included in PR #114.

PR #136, opened 2018-02-26
Fixed spacing in DocBlock

christophherr:


lib/api/wp-customize/class.php
```diff
@@ -59,7 +59,7 @@ public function __construct( $section, $args ) {
 	 *
 	 * @since 1.0.0
 	 * @ignore
-	 * @access  private
+	 * @access private
```
hellofromtonya: You don't need to make the `@access` for _methods_.  The `private` keyword tells us its access setting. 

In contrast, we use the `@access` identifier for _functions_ when they are private as a way to alert a developer of its intent.  But within a class, that's not necessary.

Therefore, please review it here.

PR #138, opened 2018-02-27
Made lib\template\fragments WPCS compliant

christophherr:


lib/templates/fragments/header.php
```diff
@@ -2,41 +2,47 @@
...
  *
  * @since 1.0.0
+ *
+ * @return void
  */
 function beans_head_meta() {
-
 	?>
-	<meta charset="<?php esc_attr_e( get_bloginfo( 'charset' ) ); ?>" />
+	<meta charset="<?php esc_attr( get_bloginfo( 'charset' ) ); ?>" />
```
hellofromtonya: Actually, we need to echo here.  This is not translated text.  Therefore, we can do this:

`echo esc_attr( get_bloginfo( 'charset' ) );`

lib/templates/fragments/header.php
```diff
@@ -47,91 +53,120 @@ function beans_favicon() {
...
  * Print the header image css inline in the header.
  *
  * @since 1.0.0
+ *
+ * @return void
  */
 function beans_header_image() {
 
-	if ( ! current_theme_supports( 'custom-header' ) || ! ( $header_image = get_header_image() ) || empty( $header_image ) ) {
+	$header_image = get_header_image();
```
hellofromtonya: If the theme does not support the header image, there's no need to waste the extra processing time. Why? It's going to bail out. Right?

Therefore, let's put back the order like this:

```
if ( ! current_theme_supports( 'custom-header' ) ) {
     return;
}

$header_image = get_header_image();

if ( empty( $header_image ) ) {
     return;
}
```

Notice too that we can get rid of the `! $header_image` as it's the same as `empty`.

lib/templates/fragments/header.php
```diff
@@ -47,91 +53,120 @@ function beans_favicon() {
...
+					array(
+						'class' => 'tm-logo',
+						'src'   => $logo, // Automatically escaped.
+						'alt'   => get_bloginfo( 'name' ), // Automatically escaped.
+					)
+				);
+			} else { // @codingStandardsIgnoreLine - Generic.WhiteSpace.ScopeIndent.IncorrectExact.
 				beans_output_e( 'beans_site_title_text', get_bloginfo( 'name' ) );
-			}
+			} // @codingStandardsIgnoreLine - Generic.WhiteSpace.ScopeIndent.IncorrectExact.
```
hellofromtonya: Get into the habit of adding _why_ we are ignoring this rule.  Add the "why" after the rule:

```
// @codingStandardsIgnoreLine - Generic.WhiteSpace.ScopeIndent.IncorrectExact - Indentation is intentional to show HTML markup structure.
```

lib/templates/fragments/menu.php
```diff
@@ -29,68 +34,85 @@ function beans_primary_menu() {
...
+		)
+	);
+
+		beans_open_markup_e(
+			'beans_primary_menu_offcanvas_button_icon',
+			'i',
+			array(
+				'class' => 'uk-icon-navicon uk-margin-small-right',
+			)
+		);
```
hellofromtonya: This is another one that _could_ have all of the arguments inline. Why? The array only has one element in it.

lib/templates/fragments/post-shortcodes.php
```diff
@@ -76,24 +91,27 @@ function beans_post_meta_comments_shortcode() {
...
+		);
 	}
 
-	beans_open_markup_e( 'beans_post_meta_comments', 'a', array(
-		'href' => get_comments_link(), // Automatically escaped.
-	) );
+	beans_open_markup_e( 'beans_post_meta_comments', 'a', array( 'href' => get_comments_link() ) ); // Automatically escaped.
 
-		printf( $comment_text, (int) get_comments_number( $post->ID ) );
+		printf( $comment_text, (int) get_comments_number( $post->ID ) ); // @codingStandardsIgnoreLine - WordPress.XSS.EscapeOutput.OutputNotEscaped.
```
hellofromtonya: Add the reason why we are ignoring this rule.

lib/templates/fragments/post-shortcodes.php
```diff
@@ -103,24 +121,23 @@ function beans_post_meta_tags_shortcode() {
 		return;
 	}
 
-	printf( '%1$s%2$s', beans_output( 'beans_post_meta_tags_prefix', __( 'Tagged with: ', 'tm-beans' ) ), $tags );
-
+	printf( '%1$s%2$s', beans_output( 'beans_post_meta_tags_prefix', __( 'Tagged with: ', 'tm-beans' ) ), $tags ); // @codingStandardsIgnoreLine - WordPress.XSS.EscapeOutput.OutputNotEscaped.
```
hellofromtonya: Add the reason why we are ignoring this rule.

lib/templates/fragments/post-shortcodes.php
```diff
@@ -103,24 +121,23 @@ function beans_post_meta_tags_shortcode() {
...
-
 	$categories = get_the_category_list( ', ' );
 
 	if ( ! $categories || is_wp_error( $categories ) ) {
 		return;
 	}
 
-	printf( '%1$s%2$s', beans_output( 'beans_post_meta_categories_prefix', __( 'Filed under: ', 'tm-beans' ) ), $categories );
-
+	printf( '%1$s%2$s', beans_output( 'beans_post_meta_categories_prefix', __( 'Filed under: ', 'tm-beans' ) ), $categories ); // @codingStandardsIgnoreLine - WordPress.XSS.EscapeOutput.OutputNotEscaped.
```
hellofromtonya: Add the reason why we are ignoring this rule.

lib/templates/fragments/breadcrumb.php
```diff
@@ -72,76 +73,59 @@ function beans_breadcrumb() {
...
+	foreach ( $breadcrumbs as $breadcrumb_url => $breadcrumb ) {
 
-				beans_open_markup_e( 'beans_breadcrumb_item', 'li' );
+		// Breadcrumb items.
+		if ( count( $breadcrumbs ) - 1 !== $i ) {
+			beans_open_markup_e( 'beans_breadcrumb_item', 'li' );
 
-					beans_open_markup_e( 'beans_breadcrumb_item_link', 'a', array(
+				beans_open_markup_e(
+					'beans_breadcrumb_item_link', 'a', array(
```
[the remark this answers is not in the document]
christophherr: I split it because of the inline comment.  I'll move the comment at the end of the line.

PR #139, opened 2018-02-27
Made lib\api\html\functions.php WPCS compliant

christophherr:


lib/api/html/functions.php
```diff
@@ -99,25 +94,24 @@ function beans_remove_output( $id ) {
  *
  * @since 1.0.0
  *
- * @param string $id               A unique string used as a reference. The $id argument may contain sub-hooks(s).
- * @param string|bool $tag         The HTML tag. If set to False or empty, the markup HTML tag will be removed but
- *                                 the actions hook will be called. If set the Null, both markup HTML tag and actions
- *                                 hooks will be removed.
+ * @param string       $id               A unique string used as a reference. The $id argument may contain sub-hooks(s).
```
hellofromtonya: Whoops, the description is not inline with the others.

lib/api/html/functions.php
```diff
@@ -99,25 +94,24 @@ function beans_remove_output( $id ) {
...
+ *                                  the actions hook will be called. If set the Null, both markup HTML tag and actions
+ *                                  hooks will be removed.
  * @param string|array $attributes Optional. Query string or array of attributes. The array key defines the
  *                                 attribute name and the array value defines the attribute value. Setting
  *                                 the array value to '' will display the attribute value as empty
  *                                 (e.g. class=""). Setting it to 'false' will only display
  *                                 the attribute name (e.g. data-example). Setting it to 'null' will not
  *                                 display anything.
- * @param mixed  $var              Optional. Additional variables passed to the functions hooked to <tt>$id</tt>.
+ * @param mixed        $var              Optional. Additional variables passed to the functions hooked to <tt>$id</tt>. // @codingStandardsIgnoreLine - Squiz.Commenting.FunctionComment.ExtraParamComment.
```
hellofromtonya: 1. Whoops, the description is not inline with the others. 
2. Add the reason why we are ignoring the rule.

lib/api/html/functions.php
```diff
@@ -405,16 +388,16 @@ function beans_reset_markup( $id ) {
...
+ * @param string       $new_id     A unique string used as a reference. The $id argument may contain sub-hook(s).
+ * @param string       $tag        The HTML wrap tag.
  * @param string|array $attributes Optional. Query string or array of attributes. The array key defines the
  *                                 attribute name and the array value define the attribute value. Setting
  *                                 the array value to '' will display the attribute value as empty
  *                                 (e.g. class=""). Setting it to 'false' will only display
  *                                 the attribute name (e.g. data-example). Setting it to 'null' will not
  *                                 display anything.
- * @param mixed  $var              Additional variables passed to the functions hooked to <tt>$id</tt>.
+ * @param mixed        $var        Additional variables passed to the functions hooked to <tt>$id</tt>. // @codingStandardsIgnoreLine - Squiz.Commenting.FunctionComment.ExtraParamComment.
```
hellofromtonya: On line 405, remove the empty line.

lib/api/html/functions.php
```diff
@@ -20,13 +22,12 @@
  *
  * @param string $id     A unique string used as a reference. The $id argument may contain sub-hook(s).
  * @param string $output Content to output.
- * @param mixed  $var    Additional variables passed to the functions hooked to <tt>$id</tt>.
+ * @param mixed  $var    Additional variables passed to the functions hooked to <tt>$id</tt>. // @codingStandardsIgnoreLine - Squiz.Commenting.FunctionComment.ExtraParamComment.
```
GaryJones: Why is this needed?

hellofromtonya: Thierry added that for the automatically generated documentation, i.e. to identify that you can pass more augments to the function.  We use `func_get_args()` to grab all of the arguments.

We've standardized on identifying this feature as an additional parameter. But one could easily argue that this is misleading. Why? Because we can pass 1 or more additional arguments.  Right? Using `get_func_args()` means we are not limited to just 1.  

Reading the parameter list in the DocBlock, it could misleading.

Another approach then is to standardize on a note in the DocBlock itself, alerting the developer that s/he can pass additional arguments.

@GaryJones What do you think?  Which approach is more clear to you?

GaryJones: I'm not worried about the existing parameter, but about the additional `// @codingStandardsIgnoreLine - Squiz.Commenting.FunctionComment.ExtraParamComment.` that's added on in this diff.

GaryJones: I'd first question why there are apparently so many functions that use `func_get_args()` that there is a need to standardize - that appears as a code smell to me.

hellofromtonya: >Also, it's using the deprecated @codingStandards... format.
We are using this because Beans 1.x has a requirement for PHP 5.2.  That's defined in our `phpcs.xml.dist` file.  With that requirement, the sniffer is not installing the latest version.  Therefore, until we bump up the PHP version (which happens in Beans 2), we are unable to use the new method of ignoring rules.  

>I'd first question why there are apparently so many functions that use func_get_args() that there is a need to standardize - that appears as a code smell to me.
The intent of `func_get_args()` in Beans is to protect maximum flexibility to the developer, meaning s/he can pass additional arguments that get filtered for determining what should be filtered.

But your note is valid, as it needs part of the audit to ensure that each one is actually required. Is there a valid use or edge case that warrants it?  Are the args being filtered?

I'll open up an issue to ensure we are looking at each one on a case-by-case basis.

GaryJones: > We are using this because Beans 1.x has a requirement for PHP 5.2. That's defined in our phpcs.xml.dist file. With that requirement, the sniffer is not installing the latest version.

I've opened #144 to discuss it further.

hellofromtonya: To close the loop here, Issue #144 was resolved with PR #145.  We have now upgraded our phpcs ignore syntax thanks to @GaryJones.

@christophherr you will want to use the new syntax of:

```
// phpcs:ignore Rule.To.Be.Ignored, Another.Rule.To.Be.Ignored -- Reason why we are ignoring this rule(s)
```

or more specifically here:

```
// phpcs:ignore Squiz.Commenting.FunctionComment.ExtraParamComment -- The extra parameter is intentional for documentation.
```

hellofromtonya:
Closing this pull request, as it has conflicts and issues.  @christophherr will open a new one.

PR #145, opened 2018-02-28
Upgrade PHPCS/WPCS

### LINKED ISSUE #144: Use latest PHP_CodeSniffer
Just because Beans is made to work (for _consumers_) at PHP 5.2, doesn't mean that development tools need that very low restriction.

PHPCS 3.* requires PHP 5.4, so it's more than fine to expect any _contributor_ to Beans to have that to be able to run the PHPCS checks.

PHPCS 3.* is a rewrite that uses namespaces etc. but also fixes _many_ bugs and has many improvements. Your other PHPCS-related dependencies all work with PHPCS 3.*.

You should be explicit about requiring it the `composer.json` (it's not there at all, which is why it only installs the 2.9 branch), and optionally specify a `php` version of 5.4 (or higher) in the `require-dev` section. 

FWIW, Genesis has PHP 5.6 or 7.* as the require-dev version, and we explicitly require PHPCS 3.2:

screenshot 2018-02-28 08 16 52

You can still set PHPCompatibility to check for PHP 5.2 compatibility for the main code like you already specify that PHP 5.6 (or higher) is needed for the `tests/` directory.

hellofromtonya:
Credit to @GaryJones - This PR upgrades our PHP sniffers to use the latest version, thereby allowing us to remove the deprecated `// @coding...` ignore declarations.

It standardizes our ignore syntax to be:

```
// phpcs:ignore Rule.To.Be.Ignored, Another.Rule.To.Be.Ignored -- Reason why we are ignoring this rule(s)
```

Closes #144

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,13 +41,15 @@
     "roave/security-advisories": "dev-master"
   },
   "require-dev": {
-    "dealerdirect/phpcodesniffer-composer-installer": "^0.4.3",
-    "sirbrillig/phpcs-variable-analysis": "^2.0",
-    "wimg/php-compatibility": "^8.0",
-    "wp-coding-standards/wpcs": "^0.14.0",
-    "phpunit/phpunit": "~4.8 || ~5.7.9",
-    "brain/monkey": "^2.0",
+    "php": "^5.6|^7",
+    "brain/monkey": "^2.2",
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.4.4",
     "mikey179/vfsStream": "^1.6",
+    "phpunit/phpunit": "~4.8 || ~5.7.9",
+    "sirbrillig/phpcs-variable-analysis": "^2.0",
+    "squizlabs/php_codesniffer": "^3.3",
+    "wimg/php-compatibility": "^8.0",
+    "wp-coding-standards/wpcs": "^0.14.1",
     "xwp/wp-dev-lib": "^1.0.1"
   },
   "config": {
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
     "brain/monkey": "^2.2",
     "dealerdirect/phpcodesniffer-composer-installer": "^0.4.4",
     "mikey179/vfsStream": "^1.6",
-    "phpunit/phpunit": "~4.8 || ~5.7.9",
+    "phpunit/phpunit": "~5.7.9",
     "sirbrillig/phpcs-variable-analysis": "^2.0",
     "squizlabs/php_codesniffer": "^3.2",
     "wimg/php-compatibility": "^8.0",
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
     "mikey179/vfsStream": "^1.6",
     "phpunit/phpunit": "~4.8 || ~5.7.9",
     "sirbrillig/phpcs-variable-analysis": "^2.0",
-    "squizlabs/php_codesniffer": "^3.3",
+    "squizlabs/php_codesniffer": "^3.2",
     "wimg/php-compatibility": "^8.0",
     "wp-coding-standards/wpcs": "^0.14.1",
     "xwp/wp-dev-lib": "^1.0.1"
--- a/lib/api/init.php
+++ b/lib/api/init.php
@@ -3,7 +3,7 @@
  *
  * Load the API components.
  *
- * @since 1.5.0
+ * @since   1.5.0
  *
  * @package Beans\Framework\API
  */
@@ -18,7 +18,7 @@ define( 'BEANS_API', true );
 
 // Mode.
 if ( ! defined( 'SCRIPT_DEBUG' ) ) {
-	define( 'SCRIPT_DEBUG', false ); // @codingStandardsIgnoreLine - WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedConstantFound.
+	define( 'SCRIPT_DEBUG', false ); // @phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedConstantFound -- Valid use case as we need it defined.
 }
 
 // Assets.
--- a/lib/api/init.php
+++ b/lib/api/init.php
@@ -3,7 +3,7 @@
  *
  * Load the API components.
  *
- * @since   1.5.0
+ * @since 1.5.0
  *
  * @package Beans\Framework\API
  */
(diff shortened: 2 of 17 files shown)


hellofromtonya:
@GaryJones Please do a formal code review for this PR. If it's acceptable, go ahead and _approve_ it.  Else, note what needs to change.

P.S. `_Beans_WP_Customize_Control` will be moved to a separate file in a different PR.

composer.json
```diff
@@ -41,13 +41,15 @@
...
   },
   "require-dev": {
-    "dealerdirect/phpcodesniffer-composer-installer": "^0.4.3",
+    "php": "^5.6|^7",
+    "brain/monkey": "^2.2",
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.4.4",
+    "mikey179/vfsStream": "^1.6",
+    "phpunit/phpunit": "~4.8 || ~5.7.9",
     "sirbrillig/phpcs-variable-analysis": "^2.0",
+    "squizlabs/php_codesniffer": "^3.3",
```
GaryJones: PHPCS 3.3 has not been released yet, and in any case, the new annotations came out in 3.2.0, so `^3.2` is the suitable constraint here, unless there's something in 3.3 that is critically required.

composer.json
```diff
@@ -41,13 +41,15 @@
     "roave/security-advisories": "dev-master"
   },
   "require-dev": {
-    "dealerdirect/phpcodesniffer-composer-installer": "^0.4.3",
+    "php": "^5.6|^7",
+    "brain/monkey": "^2.2",
+    "dealerdirect/phpcodesniffer-composer-installer": "^0.4.4",
+    "mikey179/vfsStream": "^1.6",
+    "phpunit/phpunit": "~4.8 || ~5.7.9",
```
GaryJones: If PHP 5.6 is minimum, then I think that means PHPUnit 4.* is out.

composer.json
```diff
@@ -41,13 +41,15 @@
     "roave/security-advisories": "dev-master"
   },
   "require-dev": {
-    "dealerdirect/phpcodesniffer-composer-installer": "^0.4.3",
+    "php": "^5.6|^7",
+    "brain/monkey": "^2.2",
```
GaryJones: Something critical in 2.1 or 2.2 that's needed? (new feature, important bug fix)

lib/api/init.php
```diff
@@ -3,7 +3,7 @@
  *
  * Load the API components.
  *
- * @since 1.5.0
+ * @since   1.5.0
```
GaryJones: No need to line this up.

lib/api/init.php
```diff
@@ -18,7 +18,7 @@
 
 // Mode.
 if ( ! defined( 'SCRIPT_DEBUG' ) ) {
-	define( 'SCRIPT_DEBUG', false ); // @codingStandardsIgnoreLine - WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedConstantFound.
+	define( 'SCRIPT_DEBUG', false ); // @phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedConstantFound -- Valid use case as we need it defined.
```
GaryJones: I'm surprised WPCS doesn't already allow for this exclusions.

lib/api/utilities/functions.php
```diff
@@ -86,11 +86,11 @@ function beans_remove_dir( $dir_path ) {
 		if ( is_dir( $path ) ) {
 			beans_remove_dir( $path );
 		} else {
-			@unlink( $path ); // @codingStandardsIgnoreLine - Generic.PHP.NoSilencedErrors.Discouraged  This is a valid use case.
+			@unlink( $path ); // phpcs:disable Generic.PHP.NoSilencedErrors.Discouraged -- Valid use cases.
```
GaryJones: This uses `:disable`, rather than `:ignore`. Why?

lib/api/utilities/functions.php
```diff
@@ -86,11 +86,11 @@ function beans_remove_dir( $dir_path ) {
...
 		}
 	}
 
-	return @rmdir( $dir_path ); // @codingStandardsIgnoreLine - Generic.PHP.NoSilencedErrors.Discouraged This is a valid use case.
+	return @rmdir( $dir_path ); // phpcs:disable Generic.PHP.NoSilencedErrors.Discouraged -- Valid use cases.
```
GaryJones: As above.

lib/api/widget/functions.php
```diff
@@ -6,7 +6,7 @@
  *
  * @package Beans\Framework\API\Widgets
  *
- * @since 1.0.0
+ * @since   1.0.0
```
GaryJones: No need to align this.

composer.json
hellofromtonya: Nope, nothing critical. Per [BM changelog](https://github.com/Brain-WP/BrainMonkey/releases), the minor release is for Mockery.

composer.json
hellofromtonya: You are correct. Commit coming....

lib/api/init.php
hellofromtonya: Me too.

lib/api/widget/functions.php
```diff
@@ -25,21 +25,22 @@
  *
  * @since 1.0.0
  *
- * @param array $args {
- *     Optional. Arguments used by the widget area.
+ * @param array $args                       {
```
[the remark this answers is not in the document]
hellofromtonya: It's a formatting issue with PhpStorm that I'm working on to give us a standard one-click approach to reformat and comply.

Reviewing the [WPCS document for array documentation](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#1-1-parameters-that-are-arrays), they show this:

```
/**
 * Summary.
 *
 * Description.
 *
 * @since x.x.x 
 *
 * @param type  $var Description.
 * @param array $args {
 *     Optional. An array of arguments.
 *
 *     @type type $key Description. Default 'value'. Accepts 'value', 'value'.
 *                     (aligned with Description, if wraps to a new line)
 *     @type type $key Description.
 * }
 * @param type  $var Description.
 * @return type Description.
 */
```

lib/api/init.php
GaryJones: I've opened https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1311

hellofromtonya: Perfect. Once that rule is fixed, then we can remove the ignore above. Thank you.

PR #147, opened 2018-02-28
Replaced deprecated phpcs rule ignores for api/html/attributes

hellofromtonya:
This PR replaces the deprecated `// @coding...` phpcs and wpcs rule ignores.

### DIFF
--- a/tests/phpunit/integration/api/html/beansAddAttribute.php
+++ b/tests/phpunit/integration/api/html/beansAddAttribute.php
@@ -36,7 +36,7 @@ class Tests_BeansAddAttribute extends HTML_Test_Case {
 			$this->assertSame( 10, has_filter( $hook, array( $attributes, 'add' ), 10 ) );
 			$expected              = $markup_attributes;
 			$expected['data-test'] = '';
-			$this->assertSame( $expected, apply_filters( $hook, $markup_attributes ) ); // @codingStandardsIgnoreLine - WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound  The hook's name is in the value.
+			$this->assertSame( $expected, apply_filters( $hook, $markup_attributes ) ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound -- The hook's name is in the value.
 
 			// Clean up.
 			remove_filter( $hook, array( $attributes, 'add' ), 10 );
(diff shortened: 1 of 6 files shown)


GaryJones:
For the immediate case of fixing PHPCS violations, this PR satisfies that.

But, I have concerns about the overall testing going on here.

Let's take a look at one test:

```php
// tests/phpunit/unit/api/html/beansAddAttribute.php

class Tests_BeansAddAttribute extends HTML_Test_Case {

	/**
	 * Test beans_add_attribute() should add an empty value when no value is given.
	 */
	public function test_should_add_empty_value_when_no_value_given() {

		foreach ( static::$test_markup as $beans_id => $markup ) {
			$markup_attributes = isset( $markup['attributes'] ) ? $markup['attributes'] : array();
			$attributes        = beans_add_attribute( $beans_id, 'data-test', '' );
			$hook              = $beans_id . '_attributes';

			// Set up the WordPress simulator.
			Monkey\Filters\expectApplied( $hook )
				->once()
				->with( $markup_attributes )
				->andReturn( $attributes->add( $markup_attributes ) );

			// Run the tests.
			$this->assertTrue( has_filter( $hook, array( $attributes, 'add' ), 10 ) );
			$expected              = $markup_attributes;
			$expected['data-test'] = '';
			$this->assertSame( $expected, apply_filters( $hook, $markup_attributes ) );

			// Clean up.
			remove_filter( $hook, array( $attributes, 'add' ), 10 );
		}
	}
    ... more test methods ...
}
```

So, we're testing `beans_add_attribute()`, which looks like this:

```php
function beans_add_attribute( $id, $attribute, $value ) {
	$attribute = new _Beans_Attribute( $id, $attribute, $value );

	return $attribute->init( 'add' );
}
```

Thoughts:

The test isn't a unit test, it's an integration test. Why? Because this one function under test is referencing another part of Beans (`_BeansAttribute` class). In a unit test, that class would be mocked, expecting the given constructor args and that `init` method was called once with `'add'` as the arg. Nothing more, because that's literally all the function does.

_At this point, I notice that the integration test in the parallel file appears to be exactly the same as the unit test._

As an integration test, it could go further, but right now, the logic is complicated such that it has a stack like:

- `beans_add_attribute()`
- `_Beans_Attribute::_construct()`
- `Beans_Attribute::init()`
- `beans_add_filter()`
- (Either WP's `add_filter()` or) `_beans_add_anonymous_filter()`
- `_Beans_Anonymous_Filters::_construct()`
- WPs `add_filter()`.

That's a lot of levels to test. The test code in my first snippet is relying on a filter being added 5 levels down.  That's insane.

The tests I found also don't cover every path, such as when the callback is not callable.

I'd rather see proper unit tests, and integration tests that only go to a depth of one or two and mock things, even Beans references, after that. You can still add unit tests and integration tests to what is the middle of the above stack, and check the integration a little deeper from there.

The fact that there is a stack like that (function => instantiated class => method => function => function => instantiated class => function), suggests there's a problem of _complicated_ code, even if the _complexity_ of each method and function is relatively small.

hellofromtonya:
@GaryJones Your concerns about the tests are dealt with in PR #149.  Let's move that conversation over to the new PR.

For this PR, let's focus it on replacing the deprecated PHPCS ignore syntax.

PR #148, opened 2018-02-28
Replaced deprecated phpcs rule ignores for Fields API.

hellofromtonya:
This PR replaces the deprecated `// @coding...` phpcs and wpcs rule ignores.

### DIFF
--- a/lib/api/fields/types/views/field-description.php
+++ b/lib/api/fields/types/views/field-description.php
@@ -11,4 +11,4 @@
 ?>
 
 <br /><a class="bs-read-more" href="#"><?php esc_html_e( 'More...', 'tm-beans' ); ?></a>
-<div class="bs-extended-content" style="display: none;"><?php echo $extended; // @codingStandardsIgnoreLine - WordPress.XSS.EscapeOutput.OutputNotEscaped - To optimize, escaping is handled in the calling function. ?></div>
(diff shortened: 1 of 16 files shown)


lib/api/fields/types/views/field-description.php
```diff
@@ -11,4 +11,4 @@
 ?>
 
 <br /><a class="bs-read-more" href="#"><?php esc_html_e( 'More...', 'tm-beans' ); ?></a>
-<div class="bs-extended-content" style="display: none;"><?php echo $extended; // @codingStandardsIgnoreLine - WordPress.XSS.EscapeOutput.OutputNotEscaped - To optimize, escaping is handled in the calling function. ?></div>
+<div class="bs-extended-content" style="display: none;"><?php echo $extended; ?></div><?php // phpcs:ignore WordPress.XSS.EscapeOutput.OutputNotEscaped -- To optimize, escaping is handled in the calling function. ?>
```
GaryJones: I'd rather see the escaping done as late as possible. Someone shouldn't need to go and find all the places where this file might be called from, to check if a variable is escaped or not.

hellofromtonya: I agree.  But in this case, doing so would have required us to call `wp_kses_post` twice instead of once.  For performance, we opted to call it once in the calling function.

Explanation:

The description can have `<!--more-->` in it.  When it does, we split that description into 2 parts: description and extended.  The original description is run through `wp_kses_post()` before it's split. We did that for performance.  

BUT I'll be running some benchmarks when I run this API through the micro-profiler.  I'm hoping that the impact is small enough that we can escape late inside of the view file. YTD.

PR #149, opened 2018-03-01
Improved the tests for HTML Attributes.

hellofromtonya:
Improved the unit and integration tests per @GaryJones concerns (see notes below which are copied from PR #147). 

1. Added individual method unit tests for `_Beans_Attribute`.
2. Add unit test for the `init()` method to test the expected behavior when the given method does not exist.
3. Removed unit tests for the public API functions. Why? Per @GaryJones notes below, these functions are more of an integration test.
4. Added unit tests for `_beans_is_html_dev_mode()`.

For the integration tests, I'm firing the filter in order to test the full system and ensure we get the expected result back.

Gary had noted concern about the stack-up of internal functions, such as `beans_add_filter()`.  Individual tests exist in the Filters API to test that function at a granular level.

For the complexity of the tests, I removed any extras, focusing the test specifically on checking what gets returned as it passed through the appropriate attribute method.

### Gary's Concerns

But, I have concerns about the overall testing going on here.

Let's take a look at one test:

```php
// tests/phpunit/unit/api/html/beansAddAttribute.php

class Tests_BeansAddAttribute extends HTML_Test_Case {

	/**
	 * Test beans_add_attribute() should add an empty value when no value is given.
	 */
	public function test_should_add_empty_value_when_no_value_given() {

		foreach ( static::$test_markup as $beans_id => $markup ) {
			$markup_attributes = isset( $markup['attributes'] ) ? $markup['attributes'] : array();
			$attributes        = beans_add_attribute( $beans_id, 'data-test', '' );
			$hook              = $beans_id . '_attributes';

			// Set up the WordPress simulator.
			Monkey\Filters\expectApplied( $hook )
				->once()
				->with( $markup_attributes )
				->andReturn( $attributes->add( $markup_attributes ) );

			// Run the tests.
			$this->assertTrue( has_filter( $hook, array( $attributes, 'add' ), 10 ) );
			$expected              = $markup_attributes;
			$expected['data-test'] = '';
			$this->assertSame( $expected, apply_filters( $hook, $markup_attributes ) );

			// Clean up.
			remove_filter( $hook, array( $attributes, 'add' ), 10 );
		}
	}
    ... more test methods ...
}
```

So, we're testing `beans_add_attribute()`, which looks like this:

```php
function beans_add_attribute( $id, $attribute, $value ) {
	$attribute = new _Beans_Attribute( $id, $attribute, $value );

	return $attribute->init( 'add' );
}
```

Thoughts:

The test isn't a unit test, it's an integration test. Why? Because this one function under test is referencing another part of Beans (`_BeansAttribute` class). In a unit test, that class would be mocked, expecting the given constructor args and that `init` method was called once with `'add'` as the arg. Nothing more, because that's literally all the function does.

_At this point, I notice that the integration test in the parallel file appears to be exactly the same as the unit test._

As an integration test, it could go further, but right now, the logic is complicated such that it has a stack like:

- `beans_add_attribute()`
- `_Beans_Attribute::_construct()`
- `Beans_Attribute::init()`
- `beans_add_filter()`
- (Either WP's `add_filter()` or) `_beans_add_anonymous_filter()`
- `_Beans_Anonymous_Filters::_construct()`
- WPs `add_filter()`.

That's a lot of levels to test. The test code in my first snippet is relying on a filter being added 5 levels down.  That's insane.

The tests I found also don't cover every path, such as when the callback is not callable.

I'd rather see proper unit tests, and integration tests that only go to a depth of one or two and mock things, even Beans references, after that. You can still add unit tests and integration tests to what is the middle of the above stack, and check the integration a little deeper from there.

The fact that there is a stack like that (function => instantiated class => method => function => function => instantiated class => function), suggests there's a problem of _complicated_ code, even if the _complexity_ of each method and function is relatively small.


hellofromtonya:
>That's a lot of levels to test. The test code in my first snippet is relying on a filter being added 5 levels down. That's insane.

In a full integrated, systems test, we are testing that all of the functions behave as we expect.  Just like when we use `get_option()` or `get_the_ID()` within one of our functions, we are not targeting the test at that specific used function.  Rather, we are interested in testing that the function which uses those functions behaves as we expect.

For example, `beans_remove_attribute()` is dependent upon `_Beans_Attribute` and then it's methods `init` and `remove`.  Within the `init` method, it uses `beans_add_filter()`. 

Each of those methods and functions have their own suite of tests.  But when testing `beans_remove_attribute()`, we are after the full system result.  

What happens when we fire that function with a set of inputs?  Do we get what we expect?

It does stack up.  All integrated tests will stack up.  That's the point of the integrated suite of tests, as we are ensuring the system works with everything plugged in and running as normal under the given conditions.

hellofromtonya:
>The test isn't a unit test, it's an integration test. Why?

Agreed.  I split them out.

hellofromtonya:
>The tests I found also don't cover every path, such as when the callback is not callable.

Yup, I missed that one.  I think I got all of the different scenarios.

hellofromtonya:
BTW @GaryJones - A design note for you.

The `beans_add_filter()` exists today as the means to register an anonymous function.  Closures are available in PHP 5.2

But, in Beans 2.0, the Filters API is getting a big upgrade.  It will get all of the functionality currently available in the Actions API:

- modify one of the parameters, such as change the priority or callback, without having to first do a remove and then add.
- reset back to the original configuration after a modification.
- replace or overwrite one or more of the parameters.
- remove a filter/action.

What's cool though is the above functionality is load agnostic. That means you can do all of your remove logic before it's ever added.  That gets rid of timing issues which drive people crazy (me included).

tests/phpunit/integration/api/html/beansAddAttribute.php
```diff
@@ -23,49 +23,54 @@
...
+	/**
+	 * Test beans_add_attribute() should add the attribute when it does not exist in the given attributes.
+	 */
+	public function test_should_add_the_attribute_when_does_not_exist() {
+
+		foreach ( static::$test_attributes as $beans_id => $markup ) {
+			$hook     = $beans_id . '_attributes';
+			$instance = beans_add_attribute( $beans_id, 'data-test', 'foo' );
+
+			// Run the full systems test by applying the filter.
```
GaryJones: This still feels like a system test (nothing is mocked), rather than an integration test (where just two bits of Beans is used and everything else is mocked). There's nothing wrong with having a different suite of tests to split up integration and system tests.

tests/phpunit/integration/api/html/beansRemoveAttribute.php
```diff
@@ -23,47 +23,103 @@
...
+
+		// Clean up.
+		remove_filter( 'beans_post_attributes', array( $instance, 'remove' ), 10 );
+	}
+
+	/**
+	 * Test beans_remove_attribute() should remove the attribute when the given value is null.
+	 */
+	public function test_should_remove_attribute_when_value_is_null() {
+		$attributes = array(
```
GaryJones: Why do these tests use arrays and foreach, instead of `@dataProvider`s?

tests/phpunit/unit/api/html/beans-attribute/add.php
```diff
@@ -0,0 +1,68 @@
...
+			// Check that the attribute does not exist before we add it.
+			$this->assertArrayNotHasKey( 'data-test', $markup['attributes'] );
+
+			// Check that the attribute is added with the given value.
+			$expected              = $markup['attributes'];
+			$expected['data-test'] = 'foo';
+			$this->assertSame( $expected, $instance->add( $markup['attributes'] ) );
+		}
+
+		// Check it with an empty array.
```
GaryJones: This should be a new test, or use a data provider.

tests/phpunit/unit/api/html/beans-attribute/init.php
```diff
@@ -0,0 +1,47 @@
...
+use Beans\Framework\Tests\Unit\API\HTML\Includes\HTML_Test_Case;
+use Brain\Monkey;
+
+require_once dirname( __DIR__ ) . '/includes/class-html-test-case.php';
+
+/**
+ * Class Tests_Beans_Attribute_Init
+ *
+ * @package Beans\Framework\Tests\Unit\API\HTML
+ * @group   unit-tests
```
GaryJones: A little redundant, as testsuites can be run individually anyway.

tests/phpunit/unit/api/html/beans-attribute/remove.php
```diff
@@ -0,0 +1,105 @@
...
+		$attributes = array(
+			'href'  => 'http://example.com/image.png',
+			'title' => 'Some cool image',
+		);
+
+		$instance = new _Beans_Attribute( 'beans_post', 'itemprop' );
+		$this->assertArrayNotHasKey( 'itemprop', $attributes );
+		$this->assertSame( $attributes, $instance->remove( $attributes ) );
+
+		$attributes = array(
```
GaryJones: Appears to be two tests in one again.

tests/phpunit/unit/api/html/beansIsHtmlDevMode.php
```diff
@@ -0,0 +1,52 @@
...
+ * @group   unit-tests
+ * @group   api
+ */
+class Tests_BeansIsHtmlDevMode extends HTML_Test_Case {
+
+	/**
+	 * Test _beans_is_html_dev_mode() should return false when the option does not exist.
+	 */
+	public function test_should_false_when_option_does_not_exist() {
+		$this->assertFalse( defined( 'BEANS_HTML_DEV_MODE' ) );
```
GaryJones: I'm not sure why this assertion is here.

tests/phpunit/unit/api/html/beansIsHtmlDevMode.php
```diff
@@ -0,0 +1,52 @@
...
+			->with( 'beans_dev_mode', false )
+			->once()
+			->andReturn( false );
+		$this->assertFalse( _beans_is_html_dev_mode() );
+	}
+
+	/**
+	 * Test _beans_is_html_dev_mode() should return the option's value.
+	 */
+	public function test_should_return_option_value() {
```
GaryJones: Are their System tests, which really populate the wp_options table in the test database, and then just call the function being tested?

lib/api/html/functions.php
```diff
@@ -611,7 +611,11 @@ function beans_remove_attribute( $id, $attribute, $value = null ) {
 /**
  * Check if development mode is enabled taking in consideration legacy constant.
  *
+ * @since  1.5.0
  * @ignore
+ * @access private
+ *
+ * @return bool
  */
 function _beans_is_html_dev_mode() {
```
GaryJones: The return paths here should probably explicitly cast to a boolean.

tests/phpunit/integration/api/html/beansAddAttribute.php
hellofromtonya: So you are advocating that we create 3 different suites of tests:

1. Unit - where we test without the dependencies via mocking/simulating those dependencies.
2. Integration - where we integrate the dependencies but do not test all the way through. 
3. Systems - where we do a full integration and test all the way through.

So in this case of the "add attribute", using these 3 set suites, we'd have;

1. Unit tests for the "add" and "init" methods.
2. Integration tests to ensure that the proper filter callback was registered, i.e. `has_filter()`.
3. Systems tests to fire the filter and test that we get the right results.

Is that what you are suggesting here?

tests/phpunit/unit/api/html/beansIsHtmlDevMode.php
hellofromtonya: There are 2 different results from this function:

1. That it passed the constant when defined, i.e. that's for legacy code.
2. Or when not defined, it goes to the database and returns the value if it exists.  If no, it returns the default.

tests/phpunit/integration/api/html/beansAddAttribute.php
hellofromtonya: In your opinion, you do not want to check all the way through in an integration test.  For example, for `beans_add_attribute()` the expectation are:

1. It creates an attribute object with the given inputs of ID, the target attribute name, and the value to add.
2. It registers the object's callback for the "add" task.

Those are the _technical_ expectations of the `beans_add_attribute()`.

But then we are leaving out the endpoint.  Right?  When we call `beans_add_attribute` in our code, what we are really expecting is that:

1. We've created an object that holds the targets, i.e. for this ID, I want to target this attribute name and add this value to it.
2. Then when the event fires (which happens in the same API), we expect it to add the value to attribute.

Where do we test that?  We could do it in the other HTML function that actually fires the event and we will indirectly by testing the final HTML that is built.  

But my approach is, go one more step right here where we are focusing on adding the attribute. 
 Make sure at the attribute level that the attribute is added for the given inputs.

Yes, we could create a third suite of tests and label them "systems".  But I think that is too much to maintain.  Why?  We'd be repeating the bulk of this code again, simply to test the final result.

My definition of integration test is full integration, with all dependencies functioning as expected.  A unit test tests a function scenarios without the dependencies.  An integration test integrates the dependencies into the test suite to ensure our target function performs per our expectations for the given scenarios.

hellofromtonya:
@GaryJones I think I need to do a better job at explaining the approach I've taken in designing the test suites. Indulge me a moment as we seek to fine-tune and improve the design.

- Unit tests simulate (mock) dependencies in order to test at a single unit.
- Integration tests run the dependencies as an _integrated unit_, treating them as a _black box_, and then testing the scenarios of expected behavior for each given set of conditions.  It focuses on behavior at a functional level and modular level.
- Systems tests run the entire system as a whole, testing the expected behavior of the entire suite of modules for a given system level expectation, such as the entire built template/web page. 

### How does an integration test differ from a systems test?

Integration tests are more micro and focused, as we get done to an API or set of modules, or even a specific function and how it interacts with the modules.  It's more granular, allowing us to ensure that a targeted functionality is performing within the integrated environment.

Systems test, on the other hand, is at the macro level.  For a theme, that would be testing that the entire template was built as we expect to be, with all of the modules and environment fully integrated.  It takes a big picture snapshot of how the entire system is working.

### The Design

We test both unit and integration, but not systems.  Why?  I've found little value in testing the entire web page.  Rather, I want to know more at a micro level how everything is working at a specific functionality.   I want to know that each function is performing as a unit but also as an integrated part of the entire system.

I don't think having a suite of system-level tests is productive.  We are more interested in knowing that individual functions and APIs are performing as we expect.

tests/phpunit/unit/api/html/beansIsHtmlDevMode.php
GaryJones: Right - but there's no test (or at least wasn't) that this function returns the value correctly when the constant is defined.

GaryJones:
I think this is where the discrepency is - these are the definitions I'm using (as added to Genesis):

- **Unit Tests** only test a single function or method in Genesis. Everything else, including other parts of Genesis, should be mocked.
- **Integration Tests** test how multiple parts of Genesis work together. Everything outside of Genesis should be mocked.
- **System Tests** test how Genesis and WordPress work together. It uses WP functions and the database.

In Genesis, we don't yet differentiate System Integration tests, but we're still building up the tests. What is currently our "System" tests are most likely System Integration tests.

The advantage of having System tests be the only testsuite that uses real WP functions is that it's the only one that requires us to boot up WP, which makes unit and integration tests considerably quicker (and available outside of a virtual environment). Here's a full run of the Genesis tests:

screenshot 2018-03-02 10 11 50

For Genesis, some of our functions are so intertwined with WP functions, that mocking them all is unrealistic or not valuable as a realistic test.

iCaspar:
FWIW, I've come to see, generally speaking, the outline of unit/integration/system as @hellofromtonya has outlined it above. That said, I think @GaryJones is right that Beans has a code complexity issue (shared with just about every other WP project I've been involved with) stemming from the WP spaghetti-and-meatballs code architecture and it's current non-object. In the above, it would be better, I think to start mocking other beans functions in the unit tests _as if_ they were part of some other class, rather than just requiring them in and letting them run. (I'm pretty sure Brain\Monkey provides for this.) If there are bugs in not-yet-tested functions, they'll throw off the unit test for the function under test in unpredictable ways.

hellofromtonya:
Respectfully, @GaryJones, your focus on the test suites is far too narrow.  Let me explain.

We have unit, integration, and system.  Each is building up to ensure our code is performing as designed and expected.  The unit tests at a micro level and we work our way up to the complete, fully integrated system at the system level, meaning a complete component, module, or web page.

### A single unit (micro level)

A unit tests that functionality performs as a expected at that unit level without dependencies.  It's our single block, if you will, to ensure at the micro-level that our code works.

### The integrated unit

Next we step up a level to _integrate_ our block with all the other blocks, meaning we integrate the dependencies.  

At the integration level, we are testing our our single functionality works as expected **when its dependencies are plugged into it**.  We want to know that this functionality performs as designed within the system.

The dependencies are treated as a _black box_. 

At this level, we are still focused at the functional level.

### System

A system suite of tests is intended to test a complete module, feature, or the entire web page. We want to know that all the components work as designed.

Look at our tests so far.  With the unit and integration, we've proved that _a function works_:

1. on its own with its dependencies mocked out
2. as intended with its dependencies plugged in but treated as a _black box_

- What is testing that all of the functionality works as a system?  
- What ensures that everything works together as intended to produce a complete module, feature, or web page?  
- How do we know that _all functions_ work together at the big picture level? 
- How do we know that we have not missed some wiring point or have some hole in our architecture?

All of these questions are answered at the system level.  

For example, `single.php` would be a system level test, as we are testing that all of the components and functionality are working together, fully integrated, to build and produce what we except for a single web page.

### Bigger Picture

Think of the suite of tests as these building blocks.  It helps us to visualize how the test suites work together, starting at the micro level and then moving all the way up to the final module, component, or web page.  We literally are testing low levels up to a completed system.

For Beans, currently, we are building our test suite following the above definitions.  Now when we get to the system level, then we may opt to build system tests to validate that all of the components are working together as designed.

hellofromtonya:
@GaryJones I restructured the dependent files, see PR #154 and PR #152, today.  This restructuring in the unit test suite allows us to monkey patch Beans functionality.  Starting today, we will mock out internal functions for the unit tests in order to achieve the _single unit_ intent.

I've taken an action item to update our Contributor Wiki.

Given our conversations, explanation of the Beans' test suite design (unit and integration), and the changes made per your request, I believe this PR can be approved and merged.

PR #150, opened 2018-03-01
Add tests for beans_get_term_meta()

### LINKED ISSUE #146: beans_get_term_meta() should return early when term_id can't be found
When `$term_id` can't be retrieved from `get_queried_posts()`, nor from `beans_get('tag_ID')`, calling `get_option( "beans_term_{$term_id}_{$field_id}", $default }` fails `ENULL` -- and `null` gets returned instead of `$default`.

iCaspar:
* Adds unit and integration tests for `beans_get_term_meta()`.
* Fixes #146

### DIFF
--- a/lib/api/fields/class.php
+++ b/lib/api/fields/class.php
@@ -98,7 +98,7 @@ final class _Beans_Fields {
 	}
 
 	/**
-	 * Standardize field to beans format.
+	 * Standadrize field to beans format.
 	 */
 	private function standardize_field( $field ) {
 
--- a/lib/api/fields/class.php
+++ b/lib/api/fields/class.php
@@ -98,7 +98,7 @@ final class _Beans_Fields {
 	}
 
 	/**
-	 * Standadrize field to beans format.
+	 * Standardize field to beans format.
 	 */
 	private function standardize_field( $field ) {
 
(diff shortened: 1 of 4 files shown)


lib/api/fields/class.php
```diff
@@ -98,7 +98,7 @@ private function add() {
 	}
 
 	/**
-	 * Standadrize field to beans format.
+	 * Standardize field to beans format.
```
hellofromtonya: This change is conflicting with the new Fields API merge that happened yesterday.  Go ahead and put it back to the original.

tests/phpunit/integration/api/term-meta/beansGetTermMeta.php
```diff
@@ -0,0 +1,83 @@
...
+
+namespace Beans\Framework\Tests\Integration\API\Post_Meta;
+
+use WP_UnitTestCase;
+
+/**
+ * Class Tests_BeansGetPostMeta
+ *
+ * @package Beans\Framework\Tests\Integration\API\Term_Meta
+ * @group   integration-tests
```
hellofromtonya: Remove this group.  We are no longer specifying `@group integration-tests`.

tests/phpunit/unit/api/term-meta/beansGetTermMeta.php
```diff
@@ -0,0 +1,108 @@
...
+namespace Beans\Framework\Tests\Unit\API\Term_Meta;
+
+use Beans\Framework\Tests\Unit\Test_Case;
+use Brain\Monkey;
+
+/**
+ * Class Tests_BeansGetTermMeta
+ *
+ * @package Beans\Framework\Tests\Unit\API\Term_Meta
+ * @group   unit-tests
```
hellofromtonya: Remove this group.  We are no longer specifying `@group unit-tests`.

PR #165, opened 2018-03-10
Improved Filters API tests.

hellofromtonya:
1. Replaced the stubs file with mocks.
2. Replaced the `_beans_add_anonymous_Filters()` unit test (not a unit test) with `_Beans_Anonymous_Filters` `_construct()` test.
3. Added default values for `_Beans_Anonymous_Filters` priority and args parameters.


lib/api/filters/functions.php
```diff
@@ -45,6 +45,8 @@ function beans_add_filter( $hook, $callback_or_value, $priority = 10, $args = 1
  * Sub-hooks must be set in square brackets as part of the filter id argument. Sub-hooks are cascaded
  * in a similar way to CSS classes. Maximum 3 sub-hooks allowed.
  *
+ * Note: You can pass additional arguments to the functions that are hooked <tt>$id</tt>.
```
christophherr: Should this be `... functions that are hooked to ...` ?
I like moving the 'extra parameter' into the 'description' part.

tests/phpunit/unit/api/filters/beans-anonymous-filters/_construct.php
```diff
@@ -0,0 +1,56 @@
...
+namespace Beans\Framework\Tests\Unit\API\Filters;
+
+use _Beans_Anonymous_Filters;
+use Beans\Framework\Tests\Unit\Test_Case;
+
+/**
+ * Class Tests_Beans_Anonymous_Filters_Construct.
+ *
+ * @package Beans\Framework\Tests\Unit\API\Filters
+ * @group   unit-tests
```
christophherr: @group   api
@group   api-filters

lib/api/filters/functions.php
hellofromtonya: >I like moving the 'extra parameter' into the 'description' part.

Me too.  The more I thought about it, it seems like a better way to document this than in the `@param` listing.  We can now standardize it using this approach.

PR #166, opened 2018-03-10
Standardized unit test file loading.

hellofromtonya:
Standardized unit test file loading, i.e. to wait until after Patchwork is setup and using the load function.

### DIFF
--- a/tests/phpunit/unit/api/image/includes/class-image-test-case.php
+++ b/tests/phpunit/unit/api/image/includes/class-image-test-case.php
@@ -140,16 +140,13 @@ abstract class Image_Test_Case extends Test_Case {
 	 * @since 1.5.0
 	 *
 	 * @param string $method_name Method name for which to gain access.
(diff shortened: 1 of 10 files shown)


tests/phpunit/unit/api/image/includes/class-image-test-case.php
```diff
@@ -133,6 +122,8 @@ private function set_up_virtual_filesystem() {
 
 		// Set up the "beans" directory's virtual filesystem.
 		$this->mock_filesystem = vfsStream::setup( 'uploads', 0755, $structure );
+		$this->images_dir      = vfsStream::url( 'uploads/beans/images' );
+		$this->images_url      = 'http:://example.com/uploads/beans/images/';
```
christophherr: Should `http::`  be `http:`  or `https:`?

hellofromtonya: Good catch, @christophherr.

PR #167, opened 2018-03-11
Added full integration tests for Fields API

hellofromtonya:
To ensure that everything works in the Fields API, this PR adds 2 full integration tests that register the fields and then render them.

This PR also improves the test suite by:

1. Adding Actions API Container resets in the Integration Test Case.
2. Adding Fields API Container resets in the Integration Test Case.
3. Moving the reflective methods to the Integration Test Case.
4. Adding label and description set up and tear down.

The resets ensure we start at a known state for each test.

### DIFF
--- a/tests/phpunit/integration/class-test-case.php
+++ b/tests/phpunit/integration/class-test-case.php
@@ -114,4 +114,42 @@ abstract class Test_Case extends WP_UnitTestCase {
 			$property->setValue( new \_Beans_Fields(), array() );
 		}
 	}
+
+	/**
+	 * Get reflective access to the private method.
+	 *
+	 * @since 1.5.0
+	 *
+	 * @param string $method_name Method name for which to gain access.
+	 * @param string $class_name  Name of the target class.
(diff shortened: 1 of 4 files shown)


hellofromtonya:
I created this PR because I was concerned about timing issues with the Fields API.  This PR makes sure that:

1. The Fields API works end-to-end.
2. Persistent memories are reset in-between tests.

tests/phpunit/integration/class-test-case.php
```diff
@@ -33,7 +33,7 @@ public function setUp() {
 		parent::setUp();
 		Monkey\setUp();
 
-		if ( $this->was_reset ) {
+		if ( ! $this->was_reset ) {
```
christophherr: Just to note, the setUp() resets now run for (all) 50 integration tests in the Fields API and 8 integration tests in the Filters API.
The tearDown() resets run for 18 integration tests in the Fields API and (the same) 8 integration tests in the Filters API

PR #169, opened 2018-03-12
Standardized extra vars as a note instead of @param.

hellofromtonya:
To eliminate confusion, we are standardized the extra `$var` arguments as a note instead of an extra `@param`.  

This change eliminates the need to ignore the PHPCS rule.  

It also makes the documentation more readable and understandable. Why? When listed as one `@param`, it gives the impression that you can only pass one additional argument.  But that's not accurate, as you pass one or more.

### DIFF
--- a/lib/api/html/functions.php
+++ b/lib/api/html/functions.php
@@ -327,8 +327,8 @@ function beans_close_markup_e( $id, $tag ) {
  * @since 1.0.0
  *
  * @param string          $id       The markup ID.
- * @param string|callback $markup   The replacement HTML tag. A callback is accepted if conditions needs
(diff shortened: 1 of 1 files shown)


lib/api/html/functions.php
```diff
@@ -321,7 +327,7 @@ function beans_close_markup_e( $id, $tag ) {
  * @since 1.0.0
  *
  * @param string          $id       The markup ID.
- * @param string|callback $markup   The replacment HTML tag. A callback is accepted if conditions needs
+ * @param string|callback $markup   The replacement HTML tag. A callback is accepted if conditions needs
```
christophherr: `... if conditions need`

PR #174, opened 2018-03-16
Added composer script to run both PHPUnit test suites

### LINKED ISSUE #173: Consider adding a composer script for running both PHPUnit test suites
We have a composer script that runs PHPCS for the framework and the tests.
It would be great to add a similar command for the PHPUnit test suites.

christophherr:
Closes #173

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,7 +67,7 @@
     ],
     "test-unit": "\"vendor/bin/phpunit\" --testsuite unit --color=always",
     "test-integration": "\"vendor/bin/phpunit\" --testsuite integration --configuration tests/phpunit/integration/phpunit.xml.dist --color=always",
-    "test": [
+    "run-tests": [
       "@test-unit",
       "@test-integration"
     ]
--- a/composer.json
+++ b/composer.json
@@ -66,6 +66,10 @@
       "@phpcs-tests"
     ],
     "test-unit": "\"vendor/bin/phpunit\" --testsuite unit --color=always",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -66,6 +66,10 @@
       "@phpcs-tests"
     ],
     "test-unit": "\"vendor/bin/phpunit\" --testsuite unit --color=always",
-    "test-integration": "\"vendor/bin/phpunit\" --testsuite integration --configuration tests/phpunit/integration/phpunit.xml.dist --color=always"
+    "test-integration": "\"vendor/bin/phpunit\" --testsuite integration --configuration tests/phpunit/integration/phpunit.xml.dist --color=always",
+    "test": [
```
hellofromtonya: Hmm, great idea.  But the word "test" might have a future task assigned to it from a package or composer.  Instead, how about if we give it a unique task name?

Maybe something like:

```
composer run-tests
```

Therefore, the script then would be `"run-tests"`.

christophherr: Good point!
Should we brand it as 
`composer beans-tests` ?

hellofromtonya: We could.  BUT it becomes less reusable at that point.  I like defining scripts that I can train my hands to run over and over again regardless of what package I'm in.

Let's go with `"run-tests"`.  Shoot, we can do the grouped sniffers too by saying: `composer run-phpcs`.

christophherr: Yes, keeping it consistent would be great.
Should I add the change to the PHPCS script or do you want a separate PR for that?

hellofromtonya: Separate PR would be good since this one is specific for tests.

PR #191, opened 2018-03-22
Fix #190 Checking for WP_Error again, with unit test updates

### LINKED ISSUE #190: Fatal error thrown by `class_beans_compiler::get_remote_content()` on WP_Error
Need to add a second check/bail for WP_Error returned in the second (https) call to `wp_remote_get()`.  Otherwise we get `Cannot use object as array` on L:179.

iCaspar:


lib/api/compiler/class-beans-compiler.php
```diff
@@ -476,6 +476,10 @@ public function get_remote_content() {
 			$request  = wp_remote_get( $fragment );
 		}
 
+		if ( is_wp_error( $request ) ) {
```
hellofromtonya: The error can only occur when we go into the if block above. Right?  Why? Because we are checking for the error before the if block.  Therefore, to avoid a double check, move this one into the if block above.

iCaspar: Right. Good call. Done.

Issue #143, opened 2018-02-27
Safety Audit


hellofromtonya:
As part of our audit, we need to validate that each instance of tasks being rendered out to the browser or saved into the database are properly sanitized.

christophherr:
`// phpcs:ignore WordPress.XSS.EscapeOutput.OutputNotEscaped -- Pending security audit.` 
was used to flag areas that need the security auditors' attention in PR #139

PR #196, opened 2018-03-23
Escape HTML API

hellofromtonya:
Escaped HTML built by the API.  Tested using this [script](https://gist.github.com/hellofromtonya/99cb29fcacbc14c5eac7aa1b902bd0c5), which you can put into your Beans child theme's functions.php file.

Updated Rule comments to note compliance.

Added tests to both the `beans_open_markup` and `beans_close_markup` test suites to ensure code built by Beans is safe.

This PR is part of the ongoing #143 Safety Audit.

### DIFF
--- a/lib/api/html/functions.php
+++ b/lib/api/html/functions.php
@@ -58,7 +58,7 @@ function beans_output( $id, $output ) {
  */
 function beans_output_e( $id, $output ) {
 	$args = func_get_args();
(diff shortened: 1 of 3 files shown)


hellofromtonya:
The design philosophy used is this:

1. The HTML that Beans builds and renders is escaped.
2. The theme or plugin developer extending Beans is responsible for properly escaping their code.

The HTML API has several hooks available such as `'_before_markup'`, `'_prepend_markup'`, `'_append_markup'`, and `'_after_markup'`, as well as rendering functions.  These hooks and functions are available for developers to add additional HTML where it needs to render.  Beans does not escape the output received back from these hooks/functions, but rather leaves that task to the developer.

PR #198, opened 2018-03-24
Mocked functions in Actions API unit tests. 

hellofromtonya:
1. This PR ensures that functions are mocked for _pure_ unit tests. 
2. In doing so, the private function unit test classes were copied to integration tests to ensure that all functions work as intended when integrated.

Also, in doing this rework, I improved `beans_replace_action()` by only merging the "replaced" and "added" action _when_ there is an added action.  We know this by the `$is_modified` flag.


hellofromtonya:
@christophherr This is a large PR.  Let me give you some help here:

1. No review is required for the new integration tests.  Why?  These are copies of the original unit tests for the private functions.
2. I added Brain\Monkey mocks in the unit tests for all used functions within the target function to be tested.  Whew, they are now pure unit tests.
3. Let's leave documentation improvements for another PR.  I left the documentation as it was from the last PR rounds. As we previously discussed, we can make a general task to further improve documentation in a separate sprint.

christophherr:
Thank you, Tonya.
1) 👍 
2) Looks great! That was a lot of work.  
3) Sounds like a good plan.

PR #199, opened 2018-03-27
Added unit tests for _Beans_Anonymous_Action

hellofromtonya:
This PR adds unit tests for the `_Beans_Anonymous_Action`.  It also renames the class to a singular noun, as the class is the blueprint for a single anonymous action.

I want to note here that I did review the use of registering the hook in the constructor.  This practice is frowned upon; however, in this case, the point of instantiating the object is register the anonymous callback to the given hook.  Therefore, it is a valid use case.

### DIFF
--- a/tests/phpunit/unit/api/actions/beans-anonymous-action/callback.php
+++ b/tests/phpunit/unit/api/actions/beans-anonymous-action/callback.php
@@ -1,0 +1,76 @@
+<?php
+/**
+ * Tests for callback() method for _Beans_Anonymous_Actions.
+ *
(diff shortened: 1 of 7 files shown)


tests/phpunit/unit/api/actions/beans-anonymous-action/callback.php
```diff
@@ -0,0 +1,76 @@
+<?php
+/**
+ * Tests for callback() method for _Beans_Anonymous_Action.
```
christophherr: `... the callback() method of ...`

PR #201, opened 2018-04-04
Fixed typo and sequence in Fields API tests' tearDown() methods.

christophherr:


hellofromtonya:
@christophherr We really don't need to do a remove action in each of the tearDowns. Why? We should reset the actions and any other container back to its starting state.  We do that now in the [Integration tests, which you can see here](https://github.com/Getbeans/Beans/blob/development/tests/phpunit/integration/class-test-case.php#L50).  This same strategy should be done in the Unit Tests too, i.e. copy the code for the methods and put into the baseline Test Case.

Why?  We add, remove, etc. during our tests.  But at tearDown, we want to reset everything to a starting state again.  We could do that in each test case or test class.  But it will get crazy to manage.  Instead, if we do it at the base Test Case, then every single time, the containers are reset without us having to worry about it.  Less code. More stable.

tests/phpunit/unit/api/fields/includes/class-fields-test-case.php
```diff
@@ -73,6 +72,8 @@ protected function tearDown() {
 			$property = $this->get_reflective_property( $property_name );
 			$property->setValue( new \_Beans_Fields(), array() );
 		}
```
hellofromtonya: Why do we need to reset the "registered" container before we do the parent's tear down?  Is there a side effect?

tests/phpunit/integration/api/fields/includes/class-fields-test-case.php
```diff
@@ -51,11 +51,11 @@ public function setUp() {
 	 * Cleans up the test environment after each test.
 	 */
 	public function tearDown() {
-		parent::tearDown();
-
 		remove_action( 'beans_field_group_label', 'beans_field_label' );
 		remove_action( 'beans_field_wrap_prepend_markup', 'beans_field_label' );
 		remove_action( 'beans_field_wrap_append_markup', 'beans_field_description' );
+
```
hellofromtonya: Why do we need to do the `remove_action` resets before we do the other tear down processes?  What's happening here that warrants rearranging the sequence?

tests/phpunit/unit/api/fields/includes/class-fields-test-case.php
```diff
@@ -73,6 +72,8 @@ protected function tearDown() {
 			$property = $this->get_reflective_property( $property_name );
 			$property->setValue( new \_Beans_Fields(), array() );
```
hellofromtonya: This is no longer needed as we are resetting the Fields Container in the Test Case.

tests/phpunit/integration/class-test-case.php
```diff
@@ -44,11 +44,11 @@ public function setUp() {
 	 * Cleans up the test environment after each test.
 	 */
 	public function tearDown() {
-		Monkey\tearDown();
-		parent::tearDown();
-
 		$this->reset_fields_container();
 		$this->reset_actions_container();
```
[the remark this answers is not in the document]
christophherr: I was just following https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#advanced-topics

>Similarly, tearDown() methods should always call parent::tearDown() as the very last line.

tests/phpunit/integration/api/fields/includes/class-fields-test-case.php
christophherr: Please see my previous comment.

tests/phpunit/integration/class-test-case.php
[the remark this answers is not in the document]
hellofromtonya: Agreed. Good job!

PR #203, opened 2018-04-04
Bug fix on post meta screen conditions not being correctly evaluated

iCaspar:
Check for a Boolean `true` at the beginning instead of waiting til the end. Otherwise, other conditions might not allow us to evaluate the more general case.

### DIFF
--- a/lib/api/post-meta/functions-admin.php
+++ b/lib/api/post-meta/functions-admin.php
@@ -103,6 +103,7 @@ function beans_register_post_meta( array $fields, $conditions, $section, $args =
  * @return bool
  */
 function _beans_is_post_meta_conditions( $conditions ) {
+
 	// If user has designated boolean true, it's always true. Nothing more to do here.
 	if ( true === $conditions ) {
 		return true;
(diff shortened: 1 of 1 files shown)


christophherr:
I can confirm that with these changes a metabox is showing on `post-new.php` for posts, pages and CPT posts when passing `true` for `$conditions` in `beans_register_post_meta`  (visual testing).

lib/api/post-meta/functions-admin.php
```diff
@@ -103,6 +103,10 @@ function beans_register_post_meta( array $fields, $conditions, $section, $args =
  * @return bool
  */
 function _beans_is_post_meta_conditions( $conditions ) {
+	// If user has designated boolean true, it's always true. Nothing more to do here.
```
hellofromtonya: Needs a new line above the comment.  Our coding standard is to have a new empty line above all code blocks.

iCaspar: Good to know. Done.

PR #204, opened 2018-04-04
Improved consistency of setUp() and tearDown() methods

christophherr:


christophherr:
The failures and errors are coming from moving `parent::tearDown()` after the resets in `class-actions-test-case.php`

PR #215, opened 2018-04-28
Refactored get_reflective_* methods

christophherr:


tests/phpunit/integration/class-test-case.php
```diff
@@ -173,4 +173,20 @@ protected function get_reflective_property( $property, $class_name ) {
...
+	 *
+	 * @since 1.5.0
+	 *
+	 * @param string $property Property name for which to gain access.
+	 * @param string $class_name Name of the target class.
+	 *
+	 * @return mixed
+	 * @throws \ReflectionException Throws an exception if property does not exist.
+	 */
+	protected function get_reflective_property_value( $property, $class_name ) {
```
hellofromtonya: This method is less effective in the base Test Case.  Why?  Some class constructs require arguments to be passed into them for instantiating an object.  

Therefore, this method needs to be part of the individual test suite.  Let's remove it from here.

tests/phpunit/integration/api/fields/includes/class-fields-test-case.php
```diff
@@ -58,21 +58,6 @@ public function tearDown() {
...
-	 * Get the value of the private or protected property.
-	 *
-	 * @since 1.5.0
-	 *
-	 * @param string $property Property name for which to gain access.
-	 *
-	 * @return mixed
-	 * @throws \ReflectionException Throws an exception if property does not exist.
-	 */
-	protected function get_reflective_property_value( $property ) {
```
hellofromtonya: I notes in the Integration Test Case that this method is not effective at the base level.  Why? Some class constructs require arguments to instantiate the object.  We have a choice here as it's only used in one spot:

1. Get rid of this method completely, moving the code back to calling test method.
2. Modify it to allow the class name to be passed in.

Let's go with option 1.

tests/phpunit/integration/api/fields/beansRegisterFields.php
```diff
@@ -39,7 +39,7 @@ public function test_should_register_fields() {
 		$this->assertTrue( beans_register_fields( $test_data['fields'], 'beans_tests', $test_data['section'] ) );
 
 		// Check what was registered.
-		$registered = $this->get_reflective_property_value( 'registered' );
+		$registered = $this->get_reflective_property_value( 'registered', '_Beans_Fields' );
```
hellofromtonya: You'll see my trail of notes concerning this one test.  Right now, this is the only test that uses this method. 
 Instead of building a method in the Test Case for this one usage, let's do this instead:

```
$registered_property = $this->get_reflective_property( 'registered', '_Beans_Fields' );
$registered = return $registered_property->getValue( new _Beans_Fields() );
```

tests/phpunit/unit/api/fields/beansRegisterFields.php
```diff
@@ -39,7 +39,7 @@ public function test_should_register_single_fields() {
 		$this->assertTrue( beans_register_fields( $test_data['fields'], 'beans_tests', $test_data['section'] ) );
 
 		// Check what was registered.
-		$registered = $this->get_reflective_property_value( 'registered' );
+		$registered = $this->get_reflective_property_value( 'registered', '_Beans_Fields' );
```
hellofromtonya: Use the same strategy from the integration test noted previously:

```
$registered_property = $this->get_reflective_property( 'registered', '_Beans_Fields' );
$registered = return $registered_property->getValue( new _Beans_Fields() );
```

tests/phpunit/unit/class-test-case.php
```diff
@@ -225,6 +225,23 @@ protected function get_reflective_property( $property, $class ) {
...
+	 *
+	 * @since 1.5.0
+	 *
+	 * @param string $property   Property name for which to gain access.
+	 * @param string $class_name Class name.
+	 *
+	 * @return mixed
+	 * @throws \ReflectionException Throws an exception if property does not exist.
+	 */
+	protected function get_reflective_property_value( $property, $class_name ) {
```
hellofromtonya: See notes in the Integration Test Case.  Remove this method.

tests/phpunit/integration/api/fields/beansRegisterFields.php
hellofromtonya: @christophherr Go ahead and fix it in the WP Customize API as part of this PR.  Then we can group all these changes together.  It'll also get rid of the merge conflict too.

christophherr:
Let me just fix the messed up test-case formatting...

PR #253, opened 2018-06-20
Improved UIkit API compiling. Added tests.

hellofromtonya:
This PR improves the UIkit API's compiling process by:

1.  Splitting it into 2 separate private methods: `compile_styles()` and `compile_scripts`.
2. Optimizing by moving the args filter after checking if there are assets to compile.  If there are no assets, there's no reason to filter the args.

It also adds a suite of tests for the compiling process.


lib/api/uikit/class-beans-uikit.php
```diff
@@ -51,23 +61,44 @@ public function compile() {
 		 */
 		$styles = apply_filters( 'beans_uikit_euqueued_styles', $this->register_less_components() );
 
+		// If there are not styles to compile, bail out.
```
christophherr: no styles

lib/api/uikit/class-beans-uikit.php
```diff
@@ -51,23 +61,44 @@ public function compile() {
...
-		 * Filter UIkit enqueued script components.
+		 * Filter UIkit style compiler arguments.
 		 *
 		 * @since 1.0.0
 		 *
-		 * @param array $components An array of UIkit script component files.
+		 * @param array $components An array of UIkit style compiler arguments.
 		 */
-		$scripts = apply_filters( 'beans_uikit_euqueued_scripts', $this->register_js_components() );
+		$args = apply_filters( 'beans_uikit_euqueued_styles_args', array() );
```
christophherr: `_enqueued_` or do we have to deprecate and then change the name?

lib/api/uikit/class-beans-uikit.php
```diff
@@ -51,23 +61,44 @@ public function compile() {
...
-		 * Filter UIkit style compiler arguments.
+		 * Filter UIkit enqueued script components.
 		 *
 		 * @since 1.0.0
 		 *
-		 * @param array $components An array of UIkit style compiler arguments.
+		 * @param array $components An array of UIkit script component files.
 		 */
-		$styles_args = apply_filters( 'beans_uikit_euqueued_styles_args', array() );
+		$scripts = apply_filters( 'beans_uikit_euqueued_scripts', $this->register_js_components() );
```
christophherr: `_enqueued_`

lib/api/uikit/class-beans-uikit.php
```diff
@@ -76,22 +107,14 @@ public function compile() {
 		 *
 		 * @param array $components An array of UIkit script compiler arguments.
 		 */
-		$scripts_args = apply_filters(
+		$args = apply_filters(
 			'beans_uikit_euqueued_scripts_args',
```
christophherr: Can we correct this typo? 
Deprecate `beans_uikit_euqueued_scripts_args` and use `beans_uikit_enqueued_scripts_args` instead?

tests/phpunit/integration/api/uikit/beans-uikit/compile.php
```diff
@@ -0,0 +1,272 @@
...
+.uk-alert-success{background:#f2fae3;color:#659f13}
+.uk-alert-warning{background:#fffceb;color:#e28327}
+.uk-alert-danger{background:#fff1f0;color:#d85030}
+.uk-alert-large{padding:20px}
+EOB;
+		$this->assertContains( $alert_css, $compiled_css );
+	}
+
+	/**
+	 * Test _Beans_Uikit::compile() should compile the styles with thje child theme.
```
christophherr: the child theme

tests/phpunit/integration/api/uikit/includes/class-uikit-test-case.php
```diff
@@ -117,13 +125,47 @@ protected function reset_globals() {
 			'themes'     => array(),
 		);
 
-		$_beans_uikit_registered_items = array(
-			'themes' => array(
+		$_beans_uikit_registered_items = [
```
christophherr: 🍰 🥇 💯 for short array syntax

tests/phpunit/unit/api/uikit/beans-uikit/compile.php
```diff
@@ -0,0 +1,150 @@
...
+ * Class Tests_BeansUikit_Compile
+ *
+ * @package Beans\Framework\Tests\Unit\API\UIkit
+ * @group   api
+ * @group   api-uikit
+ */
+class Tests_BeansUikit_Compile extends UIkit_Test_Case {
+
+	/**
+	 * Test _Beans_Uikit::compile() should not compile when there are not assets to compile.
```
christophherr: no assets

hellofromtonya:
>Do we have a test for what happens when the JS file has a syntax error?

@christophherr We don't have sniffers **yet** for JS.  We will in a later version.

lib/api/uikit/class-beans-uikit.php
hellofromtonya: I thought about that.  But the filter may be used elsewhere by plugins or themes.  It could cause headaches for people.  So I opted to just leave it.

I know...it goes against my better nature too.

lib/api/uikit/class-beans-uikit.php
hellofromtonya: see the note below.

lib/api/uikit/class-beans-uikit.php
hellofromtonya: see the note below.

tests/phpunit/integration/api/uikit/includes/class-uikit-test-case.php
hellofromtonya: Working on converting the rest over.  That would be a great first issue for someone to tackle.

christophherr: Yes. Let's open an issue.

lib/api/uikit/class-beans-uikit.php
christophherr: Oh well... it's being moved into a plugin anyway 🙈 🙊

hellofromtonya: Exactly.  We can deal with it once we separate it out.

Issue #254, opened 2018-06-20
Change the test suite to short array syntax


christophherr:
The test suite is running PHP 5.6. 
The short array syntax is available since [PHP 5.4](https://secure.php.net/manual/en/language.types.array.php)
>As of PHP 5.4 you can also use the short array syntax, which replaces array() with [].

Let's change our tests over.
(Note: This only applies to the test suite because Beans is supporting PHP 5.2 in the 1.X.X versions.)

ibes:
I would like to try this. 

Are there any places where the long form is needed and shouldn't be replaced?

If someone is faster than me - don't let me keep you from doing it.
The task for itself will be super boring - I am just interested in having a first commit in this project and am thankful for the "good first issue" label :)

hellofromtonya:
Hey @ibes,  Thank you for jumping on to contribute!  

Right now, the Beans sourcecode must be capable of running on PHP 5.2.  I know....yuck.  But all of our tests that are in the `tests/phpunit/` folders are on PHP 5.6.  Therefore, all of the files in the test suite can get updated to convert from `array()` to `[]`.

This task is arduous as you have to do the following:

1. Go through each file in the test suites.
2. Find each `array(` instance and then convert it.
3. Find the closing `)` and convert it to `]`.
4. Then repeat.
5. You can run the test suites by doing `composer run-tests`.
6. And then run the PHPCS sniffer to make sure it passes too, i.e. `composer phpcs-tests`.

One strategy is to do one API at a time.

1. Create a new branch `upgrade/name-of-api/array-syntax`.
2. Do that one API...both the integration and unit tests.
3. Run the test suite and PHPCS to validate.
4. Commit the changes.
5. Push the branch to your local repository.
6. Then create a PR.

What do you think?

ibes:
sounds good. 

I already figured out that regex and stuff would likely cause problems - but maybe I will give it try.

Good idea to separate it in multiple branches -> easy overview and if I get too bored someone else could jump in. 

I will give it a go

hellofromtonya:
>easy overview and if I get too bored someone else could jump in.
LOL.  Exactly! Do one. Submit the PR.  Go through the code review process with us.  Get merged in as a contributor. 

And then decide if you have the rhythm down to another.

Plus, you can take the lead on it, helping others to pitch in and lend a hand too.

ibes:
API tests converted (Pull Request existing or already merged)

- [ ] Mixed (files not related to certain API
[...]

PR #270, opened 2018-06-21
Updated array syntax of test related files that are not associated wi…

ibes:
…th certain api.

#254


ibes:
Wow - I still don't know how to rebase single commits, but for sure I spend some learning time with git today :)

PR #277, opened 2018-06-22
Updated array syntax of test for widget API.

ibes:
#254


ibes:
Thanks for the cheers - it's quite a boring task - but in the meantime I guess I optimized my workflow quite well. 
With standardized ready prepared command line commands, commit messages etc. :)

PR #281, opened 2018-06-23
Updated array syntax of test for template API.

ibes:
#254 

and fixed unnecessary whitespace / spaces in one test.

### DIFF
--- a/tests/phpunit/integration/api/template/beansLoadFragmentFile.php
+++ b/tests/phpunit/integration/api/template/beansLoadFragmentFile.php
@@ -26,7 +26,7 @@ class Tests_BeansLoadFragmentFile extends Test_Case {
 	 */
 	public function test_should_return_false_when_short_circuiting() {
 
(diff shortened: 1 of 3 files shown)


christophherr:
Hey @ibes,
the white space was not unnecessary. (As you can see by Travis CI failing)
The equal sign on line 63 is aligned with lines 73 and 74.

With version 1.5, all our php files are WPCS/PHPCS compliant.

You can set up your editor/IDE to use our phpcs.xml.dist or run either `composer run-phpcs`  or `composer phpcs-tests` to check for WPCS/PHPCS violations.

ibes:
Wow - I didn't think that would really be there for a reason so I didn't run the tests after that (not-)fix.

The white spaces look very wrong on that place but if that is what is needed here ...
Will try to reverse that commit

ibes:
https://github.com/Getbeans/Beans/pull/282
new try :)

PR #282, opened 2018-06-23
Updated array syntax of test for template API.

ibes:
#254


ibes:
7 of 8 tests where positive. 
I guess that Travis had a problem to perform a test on one configuration - not the test itself failed. 
I don't see a way to restart the test.

iCaspar:
I re-started the failing Travis build. We're passing now!

PR #286, opened 2018-06-24
Changed beans_updater to call the getbeans.io API using https

### LINKED ISSUE #285: beans_updater calls the getbeans.io API using http
We should change beans_updater to use https.

christophherr:
closes #285

### DIFF
--- a/lib/admin/updater.php
+++ b/lib/admin/updater.php
@@ -40,7 +40,7 @@ function beans_updater( $value ) {
 
 	// Query the Beans REST API if the transient is expired.
 	if ( empty( $data ) ) {
-		$response = wp_remote_get( 'http://www.getbeans.io/rest-api/', array( 'sslverify' => false ) );
+		$response = wp_remote_get( 'https://www.getbeans.io/rest-api/' );
 
 		// Retrieve data from the body and decode json format.
 		$data = json_decode( wp_remote_retrieve_body( $response ), true );


ibes:
good find!

Woah - just did a search on the whole project for "http://" and found around 1300 findings. 
A lot of them are in the tests. Oh, not really - just 200 are in the tests. 
About 1000 are in the vendors folder - so composer - so nothing to do about it.

83 are in the lib. A lot in the Doc Blocks for UiKit, some for references in the Codex, and quite some in the schema.org definitions. 

And in all DocBlocks it's 
```
 * @author Beans
 * @link   http://www.getbeans.io
 * @package Beans\Framework
```

Google will switch from showing https as `secure` to showing http as `unsecure` in July / the next Chrome version. 
I feel like Beans should strongly support traffic through https. 


I would suggest (and offer to do it) that all references to getbeans.org will be https://getbeans.org (which should be as easy as doing a global search and replace and then make sure that no of the tests has used the domain for a test. 


And I imagine, that schema.org definitions like the following can also be done via https
```
beans_open_markup_e(
	'beans_comment',
	'article',
	array(
		'id'        => 'div-comment-' . get_comment_ID(), // Automatically escaped.
		'class'     => 'uk-comment',
		'itemprop'  => 'comment',
		'itemscope' => 'itemscope',
		'itemtype'  => 'http://schema.org/Comment',
	)
);
```
If someone can confirme this - I would offer to change that. 


Also references to the codex could be changed to https - like this one
```
/**
 * Remove a registered widget area.
 *
 * Since a Beans widget area is using the WordPress sidebar, this function deregisters the defined
 * WordPress sidebar using
 * {@link http://codex.wordpress.org/Function_Reference/unregister_sidebar unregister_sidebar()}.
 *
 * @since 1.0.0
 *
 * @param string $id The ID of the registered widget area.
 *
 * @return void
 */
function beans_deregister_widget_area( $id ) {
	unregister_sidebar( $id );
}
```


All three domains are already https-only pages. 



In the tests there are around
213 http://
and 
104 https://

Seems like there is a good mix of both - I imagine that is a good thing. 
Not sure if tests should always include both versions?!

hellofromtonya:
Excellent points, @ibes!  Would you please create a separate issue for the above as we will need to:

1. Update all of the DocBlocks
2. Change the `http://schema.org` to `https://schema.org`.

Thank you!

PR #298, opened 2018-07-03
Cleaned up security audit messages & escaping.

hellofromtonya:
I had left 6 or so "pending security audit" messages in the codebase.  Thank you @christophherr for reminding me.  Each of those has been audited, corrected, and updated.

In reviewing, I noticed an inconsistency in our translation escaping.  Fixed to be:

1. Use `esc_html__()` when not an attribute and is being rendered out directly or as part of `beans_open_markup`, `beans_output`, etc.
2. Do not escape directly when it's an attribute.  Why?  Attributes are automatically escaped in Beans.


hellofromtonya:
@christophherr I know. I agree.  Now we have a standard, though we need to add it to the Wiki.

PR #330, opened 2018-08-27
Namespace other less classes in lessc.php

### LINKED ISSUE #329: Less compiler errors.
If the lessc class is included by a plugin before the theme inits then you get a lessc_parser class already defined fatal error. 
Although Lessc class is namespaced whoever did it forgot to namespace the other classes in that file, easy oversight.
PR incomming.

Pross:
Fixes #329


christophherr (approved):
Thank you for contributing to Beans, @Pross

PR #337, opened 2018-09-26
Prevented output of empty skip link markup.

### LINKED ISSUE #336: Stop empty skip link markup from outputting
The skip links can be filtered with `beans_skip_links_list` but I did not account for the case that the skip links are being removed. 
The code is outputting an empty link which shows as an empty grey box when tabbing.

christophherr:
closes #336.

### DIFF
--- a/lib/api/html/accessibility.php
+++ b/lib/api/html/accessibility.php
@@ -47,7 +47,11 @@ function beans_build_skip_links() {
 	 *     @type string Anchor text.
 	 * }
 	 */
-	$skip_links = (array) apply_filters( 'beans_skip_links_list', $skip_links );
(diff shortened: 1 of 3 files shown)


ibes (commented):
Why did you get rid of the `(array)`?

besides of that, this would also be the solution I found.

ibes:
Should there be tests added to check for this behaviour?

christophherr:
Arguably, keeping the (array) would prevent `beans_output_skip_links` from ever fataling.
The downside is that if you want to shortcircuit the skip links with the filter, you can only use `__return_null`. Without the casting, you can use `__return_false` or `__return_null` in the filter.

I've no strong preference either way.

Yes, tests should be added to `beansBuildSkipLinks.php` and a few new tests written for `beans_output_skip_links`

PR #340, opened 2018-10-09
[WIP] Fix - WPCS 1.10 will cause errors and warnings

lrobi2014:
This PR includes a lot of class and file renaming.


christophherr:
Hi @lrobi2014,
Thank you for taking on this issue :)

How are you running phpcs?
Our phpcs.xml.dist is ignoring file and class names, you should not receive any errors/warnings about those. Or in other words, we are not going to rename them.
As a matter of convention, file or class names starting with `_` are considered private, i.e. only for internal use. 

We are ignoring the files in `lib/api/compiler/vendors` because they are from external sources.

lrobi2014:
Hi @christophherr,

Erm, maybe I made a mistake in the way I was running it 😓 

I use, for example:  `phpcs --extensions=php --standard=Wordpress-Core lib/api-v`

I can revert the renaming?

christophherr:
Hi @lrobi2014,

That looks like you are running a globally installed version of phpcs.

If possible, you should only run the versions that a project is using.
To do that, you have to run `composer install` in the directory with the Beans theme. 
Then you can also use our composer scripts, e.g. `composer run-phpcs` or `composer run-tests`.
(Take a look in composer.json to see the available scripts)

Tonya wrote up instructions in our Wiki: https://github.com/Getbeans/Beans/wiki/Contributors-Quick-Start

If you have any questions, you can also contact me in the [Beans Slack group](https://join.slack.com/t/getbeans/shared_invite/enQtMjcxMjM3NDQ2NjI0LWFjNjBjZDA3OWY2YzgyM2JhODM0Y2YxNmRjOTI1MjhlMTQ4ZWNmZDViZjFmOTdmOTc5MGU0MzYzNmM4YmE5YjI).

Issue #344, opened 2018-10-13
Fix failing integration test in getRemoteContent.php


christophherr:
[This test](https://github.com/Getbeans/Beans/blob/development/tests/phpunit/integration/api/compiler/beans-compiler/getRemoteContent.php#L92) is failing with the following error:
```[php]
class WP_Error#2876 (2) {

  public $errors =>
  array(1) {
    'http_request_failed' =>
    array(1) {
      [0] =>
      string(134) "cURL error 60: server certificate verification failed. CAfile: /tmp/wordpress/src/wp-includes/certificates/ca-bundle.crt CRLfile: none"
    }
  }

  public $error_data =>
  array(0) {
  }
}
```

After talking with @hellofromtonya, we are going to skip this test for now but we have to figure out what is causing the error and how to fix it.

christophherr:
Closed by #347

PR #345, opened 2018-10-13
Skipped test failing because of cURL error

christophherr:
Fixed WPCS errors to pass Sniffer.
see: #344

### DIFF
--- a/lib/api/image/class-beans-image-editor.php
+++ b/lib/api/image/class-beans-image-editor.php
@@ -1,4 +1,4 @@
-g<?php
+<?php
 /**
  * This class provides the means to edit an image.
  *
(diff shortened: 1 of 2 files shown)


iCaspar (approved):

iCaspar:
@christophherr I'm not sure what this test actually proves, other than that `wp_remote_get()` sometimes fails. As such, if we have or can write a unit test that covers the internals of the method, such that if it fails we can prove that it does so gracefully, I'd be in favor of just removing this (and probably the one for http as well).

christophherr:
Thank you, @iCaspar.
It's one of the tests that confirms what we know from the unit tests but with actual content...

PR #347, opened 2018-10-28
Fixed failing integration test in getRemoteContent.php

Tsquare17:
Problem may be related to: https://github.com/travis-ci/travis-ci/issues/6339

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ cache:
     vendor
 
 php:
-  - 5.6
+  - 5.6.37
   - 7.2
 
 env:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -7,8 +7,8 @@ cache:
     vendor
 
 php:
-  - 5.6
-  - 7.2
+  - 5.6.37
+  - 7.2.6
```
christophherr: Using `7.2` works, so let's keep it at that version for now.
```suggestion
  - 7.2
```

christophherr (changes requested):
Great find, @Tsquare17!
Could you please remove the "skip" marker from [this test](https://github.com/Getbeans/Beans/blob/development/tests/phpunit/integration/api/compiler/beans-compiler/getRemoteContent.php#L99), too?

christophherr (approved):
Thank you for contributing to Beans, @Tsquare17!

PR #351, opened 2018-11-08
Fixed skip links being output before opening body tag.

Tsquare17:


christophherr (approved):
Good catch! 
Thank you, @Tsquare17.

PR #353, opened 2018-12-02
Fixed _Beans_Image_Editor::run() can return path instead of url

Tsquare17:
If the image doesn't exist, and image editing fails, the absolute path is returned instead of the src.

### DIFF
--- a/lib/api/image/class-beans-image-editor.php
+++ b/lib/api/image/class-beans-image-editor.php
@@ -83,7 +83,7 @@ final class _Beans_Image_Editor {
 			return $this->get_image_info( beans_path_to_url( $this->rebuilt_path ), true );
 		}
 
-		return $this->get_image_info( $this->src );
+		return $this->get_image_info( beans_path_to_url( $this->src ) );
 	}
 
 	/**
--- a/tests/phpunit/integration/api/image/beans-image-editor/run.php
+++ b/tests/phpunit/integration/api/image/beans-image-editor/run.php
@@ -72,7 +72,7 @@ class Tests_BeansImageEditor_Run extends Image_Test_Case {
 	 * Test _Beans_Image_Editor::run() should return original src when the image does not exist.
 	 */
 	public function test_should_return_original_src_when_no_image() {
-		$src    = 'path/does/not/exist/image.jpg';
+		$src    = 'http://example.org/path/does/not/exist/image.jpg';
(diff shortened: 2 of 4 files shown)


christophherr:
Hey @Tsquare17,
What does this change fix?
What is the benefit of this change? 
What could break from this change?

Tsquare17:
@christophherr,
So, I came across this when doing some testing locally. I created a post and uploaded an image. But, when viewing the site, the images were broken, having an src that contained the absolute path to the original image. 

I found the problem was caused actually caused by a missing php module on my system, but what happens is that if an error occurs attempting to edit the image, instead of returning:

`return $this->get_image_info( beans_path_to_url( $this->rebuilt_path ), true )`

with the src to the edited image, beans returns:

`$this->get_image_info( $this->src )`

which doesn't use beans_path_to_url, resulting in an image with src containing the path.

for example:
`/var/www/html/testing/wp-content/uploads/2018/11/image.jpg`
instead of:
`http://testing.com/wp-content/uploads/2018/12/image.jpg`

christophherr:
Thank you for providing more details 👍 
Could you check if we changed this in the 1.5 update or if it has been like this in 1.4?
Could it make sense to check if the PHP extension/module is available and throw an exception instead?

Tsquare17:
It looks like it was introduced when refactoring after 1.5. [Here is the commit.](https://github.com/Getbeans/Beans/commit/d231c887c0865b0d46bd40948dc1c89371412fa6#diff-9150ae87d7127e4f15357f3b3fd7be20)

Do you mean to throw an exception rather than returning the original src?

christophherr:
Thanks for the detective work, @Tsquare17 👍 

> Do you mean to throw an exception rather than returning the original src?

You said that you were missing a PHP module and with the module things were working. 
I was wondering what is better; to run the extra function or let people know that a common/needed php extension/module is missing. 
Was it a common module that was missing or something out of the ordinary?

Tsquare17:
I believe it was php-gd that was missing. People aren't likely to run into that module missing unless they're setting up their own servers and neglect to install it, like someone here... 

Failing gracefully would be the easy fix, and what I think was initially intended, but if you think a notice would be more appropriate, then we can certainly go that route. I'm not sure what the best way to go about that would be, but if you could help me in the right direction, I can give it a shot.

